PR #50, opened 2026-04-25
feat: proactive large-table warning in query editor (closes #14)

### LINKED ISSUE #14: Results pagination for large result sets
The client currently receives every row the query returns. A million-row table blows memory.

**Proposal**
- Server: accept a `limit`/`offset` pair or a row cap.
- Client: render a "Load more" button or pager; show total row estimate.
- Consider keyset pagination for stable ordering.

eugene-mobile:
## Summary

- Amber warning banner appears in the query editor (between toolbar and code area) as soon as the query targets a large table without a `LIMIT`
- Banner disappears instantly as the user types `LIMIT` — no need to run the query first
- Uses the schema row counts already loaded in the sidebar, so it's zero-cost (no extra API calls)
- Detection is intentionally conservative: only fires on `SELECT` against a single known table name; joins, subqueries, CTEs, aggregations, and `COUNT(*)` queries are unaffected
- Threshold: 5,000 rows

Closes #14

## Test plan

- [ ] Type `SELECT * FROM large_table` (>5k rows, no LIMIT) → banner appears immediately
- [ ] Add `LIMIT 100` to the query → banner disappears instantly
- [ ] Type `SELECT COUNT(*) FROM large_table` → no banner (no LIMIT needed for aggregations)
- [ ] Type `SELECT * FROM large_table LIMIT 100` from the start → no banner
- [ ] Queries against tables with ≤ 5,000 rows → no banner

### DIFF
--- a/src/components/QueryEditor.tsx
+++ b/src/components/QueryEditor.tsx
@@ -1,7 +1,22 @@
 import { useEffect, useRef, useState, CSSProperties } from 'react';
 import type { Theme } from '../theme';
+import type { SchemaData } from '../api';
 import type { HistoryEntry } from '../queryHistory';
 import type { SavedQuery } from '../savedQueries';
+
+const LARGE_ROW_THRESHOLD = 5_000;
+
+function detectLargeTable(sql: string, schemaData: SchemaData | undefined): { table: string; rows: number } | null {
+  if (!schemaData || !sql.trim()) return null;
+  if (!/\bSELECT\b/i.test(sql)) return null;
+  if (/\bLIMIT\b/i.test(sql)) return null;
+  const match = sql.match(/\bFROM\s+`?(\w+)`?/i);
+  if (!match) return null;
+  const name = match[1].toLowerCase();
+  const table = schemaData.tables.find(t => t.name.toLowerCase() === name);
+  if (!table || table.rows <= LARGE_ROW_THRESHOLD) return null;
+  return { table: table.name, rows: table.rows };
+}
 
 interface QueryEditorProps {
   value: string;
@@ -9,6 +24,7 @@ interface QueryEditorProps {
   onRun: () => void;
   isRunning: boolean;
   activeSchema: string;
+  schemaData?: SchemaData;
   history?: HistoryEntry[];
   onReopenHistory?: (entry: HistoryEntry) => void;
   onDeleteHistoryEntry?: (id: string) => void;
@@ -49,7 +65,7 @@ const ToolBtn = ({ title, onClick, active, children, t }: { title: string; onCli
 );
 
 export function QueryEditor({
-  value, onChange, onRun, isRunning, activeSchema,
+  value, onChange, onRun, isRunning, activeSchema, schemaData,
   history = [], onReopenHistory, onDeleteHistoryEntry, onClearHistory,
   savedQueries = [], onSaveQuery, onDeleteSavedQuery, onRenameSavedQuery, onReopenSavedQuery,
   t,
@@ -427,6 +443,24 @@ export function QueryEditor({
           {activeSchema}
         </span>
       </div>
+      {(() => {
+        const warn = detectLargeTable(value, schemaData);
+        if (!warn) return null;
+        return (
+          <div style={{
+            display: 'flex', alignItems: 'center', gap: 8, padding: '5px 14px',
+            background: t.colorWarningBg, borderBottom: `1px solid ${t.colorWarning}40`,
+            flexShrink: 0,
+          }}>
+            <svg width="12" height="12" viewBox="0 0 24 24" fill="none" stroke={t.colorWarning} strokeWidth="2" strokeLinecap="round" strokeLinejoin="round">
+              <path d="M10.29 3.86L1.82 18a2 2 0 0 0 1.71 3h16.94a2 2 0 0 0 1.71-3L13.71 3.86a2 2 0 0 0-3.42 0z"/><line x1="12" y1="9" x2="12" y2="13"/><line x1="12" y1="17" x2="12.01" y2="17"/>
+            </svg>
+            <span style={{ fontSize: 11, color: t.colorWarning, fontFamily: '"IBM Plex Sans", sans-serif' }}>
+              <strong>{warn.table}</strong> has <strong>{warn.rows.toLocaleString()}</strong> rows — add a <code style={{ fontFamily: 'monospace' }}>LIMIT</code> to avoid loading the full table.
+            </span>
+          </div>
+        );
+      })()}
       <div style={s.editorWrap}>
         <div style={s.lineNums} aria-hidden="true">
           {value.split('\n').map((_, i) => (
--- a/src/components/QueryEditor.tsx
+++ b/src/components/QueryEditor.tsx
@@ -536,7 +536,6 @@ export function QueryEditor({
           {activeSchema}
         </span>
       </div>
-<<<<<<< HEAD
       {(() => {
         const warn = detectLargeTable(value, schemaData);
         if (!warn) return null;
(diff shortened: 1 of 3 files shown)


eugene-mobile (commented):
## Code Review

Good additions overall — the proactive large-table warning and the copy cell/row feature are both well-scoped and zero-cost. A few issues worth addressing before merge, ranging from a correctness bug to accessibility and minor style nits.

### Summary of findings

- **Bug:** `detectLargeTable` misidentifies the table when a schema-qualified name like `schema.table` is used — the regex captures the schema prefix, not the table name.
- **Fragile:** The `40` hex-opacity suffix on `t.colorWarning` works today (theme values are 6-digit hex) but is inconsistent with how other border colours are defined in `theme.ts` and will silently break if the value ever changes to `rgb()`.
- **Accessibility:** The warning banner has no `role="alert"`, so screen readers won't announce it when it appears.
- **Style:** The `{(() => { ... })()}` IIFE in JSX can be replaced with a plain variable computed before the `return`, which is more idiomatic.
- **Minor:** `menuItemStyle` is re-created on every render inside the context-menu IIFE; a module-level const avoids the churn.
- **Minor:** `contextMenu.col!` non-null assertion inside a `contextMenu.col !== null` guard — a local variable would be cleaner.
- **Deprecation:** `document.execCommand('copy')` in the clipboard fallback is deprecated; add a comment explaining why it's there.
- **Tests:** `detectLargeTable` is a pure function with several edge cases — unit tests would lock in the behaviour.

See inline comments for specifics.

src/components/QueryEditor.tsx
```diff
@@ -1,14 +1,30 @@
...
 import type { HistoryEntry } from '../queryHistory';
 import type { SavedQuery } from '../savedQueries';
 
+const LARGE_ROW_THRESHOLD = 5_000;
+
+function detectLargeTable(sql: string, schemaData: SchemaData | undefined): { table: string; rows: number } | null {
+  if (!schemaData || !sql.trim()) return null;
+  if (!/\bSELECT\b/i.test(sql)) return null;
+  if (/\bLIMIT\b/i.test(sql)) return null;
+  const match = sql.match(/\bFROM\s+`?(\w+)`?/i);
```
eugene-mobile: **Bug – schema-qualified table names are misidentified.**

The regex `\bFROM\s+\`?(\w+)\`?` captures only the first bare identifier after `FROM`. If the user writes `FROM mydb.users` or `FROM \`mydb\`.\`users\``, `match[1]` will be `mydb` (the schema), not `users` (the table), so the lookup against `schemaData.tables` silently misses and the warning never fires.

Suggested fix:
```ts
const match = sql.match(/\bFROM\s+(?:`?\w+`?\.)?`?(\w+)`?/i);
```
This skips an optional `schema.` prefix before capturing the table name.

src/components/QueryEditor.tsx
```diff
@@ -427,6 +443,24 @@ export function QueryEditor({
           {activeSchema}
         </span>
       </div>
+      {(() => {
+        const warn = detectLargeTable(value, schemaData);
+        if (!warn) return null;
+        return (
+          <div style={{
+            display: 'flex', alignItems: 'center', gap: 8, padding: '5px 14px',
+            background: t.colorWarningBg, borderBottom: `1px solid ${t.colorWarning}40`,
```
eugene-mobile, line 546: **Fragile hex-opacity suffix.**

Appending `40` to `t.colorWarning` is a hex-alpha trick that only works when the value is a 6-digit hex string. The theme already bakes opacity into similar border tokens (e.g. `colorErrorBorder: '#F0614A40'`). The warning border colour should follow the same pattern — either add a `colorWarningBorder` token to `theme.ts`, or move the opacity into the existing `colorWarning` definitions.

As-is the code works, but it is fragile and inconsistent: a future change to `rgb(…)` format would silently produce an invalid CSS value.

src/components/QueryEditor.tsx
```diff
@@ -427,6 +443,24 @@ export function QueryEditor({
           {activeSchema}
         </span>
       </div>
+      {(() => {
```
eugene-mobile: **Prefer a variable over an IIFE in JSX.**

The `{(() => { ... })()}` pattern works but is harder to scan than the idiomatic React alternative — compute the value before `return`:

```tsx
const largeTableWarn = detectLargeTable(value, schemaData);
// …
return (
  <>
    {/* … */}
    {largeTableWarn && (
      <div style={{ /* … */ }}>
        {/* … */}
      </div>
    )}
  </>
);
```

src/components/QueryEditor.tsx
```diff
@@ -427,6 +443,24 @@ export function QueryEditor({
...
+        const warn = detectLargeTable(value, schemaData);
+        if (!warn) return null;
+        return (
+          <div style={{
```
eugene-mobile: **Accessibility: add `role="alert"` to the warning banner.**

Without a live-region role, screen readers won't announce the banner when it appears dynamically. One-liner fix:

```tsx
<div role="alert" style={{ … }}>
```

`role="alert"` implies `aria-live="assertive"`; use `aria-live="polite"` if you'd prefer it not to interrupt ongoing speech.

src/components/ResultsTable.tsx
```diff
@@ -80,6 +80,24 @@ function isBoolColumn(schemaData: SchemaData | undefined, meta: ColumnMeta | und
...
+}
+
+function copyToClipboard(text: string) {
+  navigator.clipboard.writeText(text).catch(() => {
+    const ta = document.createElement('textarea');
+    ta.value = text;
+    ta.style.cssText = 'position:fixed;opacity:0';
+    document.body.appendChild(ta);
+    ta.select();
+    document.execCommand('copy');
```
eugene-mobile, line 97: **`document.execCommand` is deprecated.**

This fallback only runs when `navigator.clipboard.writeText` rejects (non-secure context or permission denied). `execCommand` is deprecated across all major browsers — it still works in practice today, but add a comment so a future reader doesn't remove it:

```ts
// Fallback for non-secure contexts where the Clipboard API is unavailable.
document.execCommand('copy');
```

src/components/ResultsTable.tsx
```diff
@@ -802,6 +845,12 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
 
       {contextMenu && (() => {
         const { target, reason } = resolveDeleteTarget(contextMenu.row, results.columnMeta);
+        const menuItemStyle: CSSProperties = {
```
eugene-mobile, line 849: **`menuItemStyle` is re-created on every render.**

This object literal sits inside `{contextMenu && (() => { … })()}` and is therefore allocated whenever `contextMenu` is truthy. Move it to a module-level constant:

```ts
const MENU_ITEM_STYLE: CSSProperties = {
  width: '100%', textAlign: 'left', padding: '6px 10px', border: 'none',
  background: 'transparent',
  cursor: 'pointer', borderRadius: 3, fontSize: 12, fontFamily: 'inherit',
  display: 'flex', alignItems: 'center', justifyContent: 'space-between', gap: 16,
};
```

Note: `color` is per-button (depends on disabled state), so omit it from the shared object and pass it inline per-button.

src/components/ResultsTable.tsx
```diff
@@ -812,6 +861,38 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
               padding: 4, fontFamily: '"IBM Plex Sans", sans-serif', fontSize: 12,
             }}
           >
+            {contextMenu.col !== null && (
+              <button
+                style={menuItemStyle}
+                onClick={() => {
+                  const meta = results.columnMeta?.find(m => m.name === contextMenu.col);
+                  copyToClipboard(cellDisplayValue(contextMenu.row[contextMenu.col!] ?? null, meta, schemaData));
```
eugene-mobile, line 870: **Prefer a local variable over the `!` assertion.**

TypeScript can't narrow `contextMenu.col` inside the callback even though we're already inside a `contextMenu.col !== null` guard. Capturing it in a local variable avoids the assertion and makes the intent clearer:

```tsx
{contextMenu.col !== null && (() => {
  const { col, row } = contextMenu; // col is string here
  return (
    <button
      onClick={() => {
        const meta = results.columnMeta?.find(m => m.name === col);
        copyToClipboard(cellDisplayValue(row[col] ?? null, meta, schemaData));
        setContextMenu(null);
      }}
    >
```

PR #51, opened 2026-04-26
feat: extend Show schema to views, procedures, and triggers (#15)

### LINKED ISSUE #15: Extend Show schema to views, procedures, functions, triggers
PR #4 added "Show schema" for base tables (and works for views via `SHOW CREATE VIEW`). Expose it via right-click on the schema browser's views/procedures/functions/triggers groups too.

**Proposal**
- Use the appropriate `SHOW CREATE …` variant for each object type.
- Reuse the existing `ShowSchemaDialog`.

eugene-mobile:
## Summary

- Views, procedures, and triggers now appear as clickable items in the schema browser sidebar (previously all three sections showed a static placeholder even when objects existed)
- Click any item to open its definition dialog; right-click for context menu "Show schema"
- Dialog title adapts to the object type: "View definition", "Procedure definition", "Trigger definition"
- "Drop table" in the context menu is now table-only (hidden for views/procedures/triggers)
- Server: /api/table-ddl accepts a new type query param and dispatches SHOW CREATE PROCEDURE or SHOW CREATE TRIGGER as appropriate

Closes #15

## Test plan

- [ ] Expand Views, click item — "View definition" dialog shows CREATE VIEW DDL
- [ ] Expand Procedures, click item — "Procedure definition" dialog shows CREATE PROCEDURE DDL
- [ ] Expand Triggers, click item — "Trigger definition" dialog shows CREATE TRIGGER DDL
- [ ] Right-click any non-table item — context menu shows "Show schema" only (no "Drop table")
- [ ] Right-click a table — context menu still shows both "Show schema" and "Drop table"
- [ ] Empty sections still show "No views / No procedures / No triggers"

### DIFF
--- a/server/src/routes/tableDdl.ts
+++ b/server/src/routes/tableDdl.ts
@@ -17,6 +17,10 @@ export const getTableDdl: RequestHandler = async (req, res) => {
     res.status(400).json({ error: 'table query param is required.' });
     return;
   }
+  if (!['table', 'view', 'procedure', 'trigger'].includes(type)) {
+    res.status(400).json({ error: `Invalid type "${type}". Must be one of: table, view, procedure, trigger.` });
+    return;
+  }
 
   try {
     const pool = getPool();
@@ -26,6 +30,7 @@ export const getTableDdl: RequestHandler = async (req, res) => {
     if (type === 'procedure') {
       sql = `SHOW CREATE PROCEDURE ${qualified}`;
     } else if (type === 'trigger') {
+      // Schema-qualified form requires MySQL 8.0+
       sql = `SHOW CREATE TRIGGER ${qualified}`;
     } else {
       // 'table' or 'view' — SHOW CREATE TABLE works for both
--- a/server/src/routes/tableDdl.ts
+++ b/server/src/routes/tableDdl.ts
@@ -7,6 +7,7 @@ const escapeIdent = (s: string) => '`' + s.replace(/`/g, '') + '`';
 export const getTableDdl: RequestHandler = async (req, res) => {
   const schema = req.query['schema'] as string | undefined;
   const table = req.query['table'] as string | undefined;
+  const type = (req.query['type'] as string | undefined) ?? 'table';
 
   if (!schema || !schema.trim()) {
     res.status(400).json({ error: 'schema query param is required.' });
@@ -20,15 +21,30 @@ export const getTableDdl: RequestHandler = async (req, res) => {
   try {
     const pool = getPool();
     const qualified = `${escapeIdent(schema)}.${escapeIdent(table)}`;
-    const [rows] = await pool.query<RowDataPacket[]>(`SHOW CREATE TABLE ${qualified}`);
+
+    let sql: string;
+    if (type === 'procedure') {
+      sql = `SHOW CREATE PROCEDURE ${qualified}`;
+    } else if (type === 'trigger') {
+      sql = `SHOW CREATE TRIGGER ${qualified}`;
+    } else {
+      // 'table' or 'view' — SHOW CREATE TABLE works for both
+      sql = `SHOW CREATE TABLE ${qualified}`;
+    }
+
+    const [rows] = await pool.query<RowDataPacket[]>(sql);
     if (rows.length === 0) {
       res.status(404).json({ error: `No DDL returned for ${qualified}.` });
       return;
     }
-    // SHOW CREATE TABLE returns [{ Table: '…', 'Create Table': '…' }] for base tables,
-    // or [{ View: '…', 'Create View': '…', character_set_client, collation_connection }] for views.
     const row = rows[0] as Record<string, unknown>;
-    const ddl = (row['Create Table'] ?? row['Create View'] ?? '') as string;
+    const ddl = (
+      row['Create Table'] ??
+      row['Create View'] ??
+      row['Create Procedure'] ??
+      row['SQL Original Statement'] ??   // SHOW CREATE TRIGGER
+      ''
+    ) as string;
     res.json({ ddl });
   } catch (err) {
     const message = err instanceof Error ? err.message : String(err);
--- a/src/api.ts
+++ b/src/api.ts
@@ -82,8 +82,9 @@ export const api = {
     return request<SchemaData>(`/api/schema?schema=${encodeURIComponent(schema)}`);
   },
 
-  tableDdl(schema: string, table: string) {
-    return request<{ ddl: string }>(`/api/table-ddl?schema=${encodeURIComponent(schema)}&table=${encodeURIComponent(table)}`);
+  tableDdl(schema: string, name: string, type?: string) {
+    const t = type ? `&type=${encodeURIComponent(type)}` : '';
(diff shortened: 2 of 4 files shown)


server/src/routes/tableDdl.ts
```diff
@@ -20,15 +21,30 @@ export const getTableDdl: RequestHandler = async (req, res) => {
   try {
     const pool = getPool();
     const qualified = `${escapeIdent(schema)}.${escapeIdent(table)}`;
-    const [rows] = await pool.query<RowDataPacket[]>(`SHOW CREATE TABLE ${qualified}`);
+
+    let sql: string;
+    if (type === 'procedure') {
+      sql = `SHOW CREATE PROCEDURE ${qualified}`;
+    } else if (type === 'trigger') {
+      sql = `SHOW CREATE TRIGGER ${qualified}`;
```
eugene-mobile, line 34: **Potential bug: `SHOW CREATE TRIGGER` may not accept a schema-qualified name.**

MySQL's documented syntax for this command is:
```sql
SHOW CREATE TRIGGER trigger_name
```
The official docs don't mention schema qualification, whereas `SHOW CREATE TABLE` and `SHOW CREATE PROCEDURE` both support `schema.name` without issue.

The generated SQL here is `SHOW CREATE TRIGGER \`schema\`.\`trigger_name\``, which works on some MySQL 8.0 versions but is not guaranteed. The safer approach is to set the schema context at connection time (or per-query via `pool.query('USE ??', [schema])` before the SHOW) rather than relying on qualification.

Please add an explicit test against a real trigger before merging.

eugene-mobile (commented):
## Code Review

Clean, well-scoped feature. The server-side dispatch, the component rename (`table` → `name`/`type`), and the `DIALOG_TITLE` lookup table are all good patterns. A few issues to address before merge:

### Findings

- **Potential bug:** `SHOW CREATE TRIGGER` may not accept a schema-qualified name — MySQL docs only document the unqualified form; worth an explicit test against a real trigger before shipping.
- **Missing server-side validation:** The `type` query param is not validated against an allowlist; unknown values silently fall through to `SHOW CREATE TABLE` rather than returning a 400.
- **Type safety gap:** `ObjectType` is defined in a UI component (`ShowSchemaDialog.tsx`) but used as a domain/API-layer concept; it should live in `api.ts`. Separately, `api.tableDdl`'s `type` parameter is typed as plain `string` rather than `ObjectType`, losing the union constraint at the call site.
- **Style (repeated):** Same IIFE-in-JSX pattern flagged in PR #50 — extract to a helper or pre-compute outside the render expression.

See inline comments for specifics.

server/src/routes/tableDdl.ts
```diff
@@ -7,6 +7,7 @@ const escapeIdent = (s: string) => '`' + s.replace(/`/g, '') + '`';
 export const getTableDdl: RequestHandler = async (req, res) => {
   const schema = req.query['schema'] as string | undefined;
   const table = req.query['table'] as string | undefined;
+  const type = (req.query['type'] as string | undefined) ?? 'table';
```
eugene-mobile, line 10: **Missing validation: unknown `type` values silently fall through.**

Any string other than `'procedure'` or `'trigger'` (e.g. a typo, or a future client bug) falls through to `SHOW CREATE TABLE` without error. This can produce a misleading success response. Add an explicit allowlist check alongside the existing `schema`/`table` guards:

```ts
const VALID_TYPES = ['table', 'view', 'procedure', 'trigger'] as const;
type ValidType = typeof VALID_TYPES[number];
const type = (req.query['type'] as string | undefined) ?? 'table';
if (!VALID_TYPES.includes(type as ValidType)) {
  res.status(400).json({ error: `Invalid type '${type}'.` });
  return;
}
```

src/components/ShowSchemaDialog.tsx
```diff
@@ -2,14 +2,24 @@ import { useEffect, useState, CSSProperties } from 'react';
 import type { Theme } from '../theme';
 import { api } from '../api';
 
+export type ObjectType = 'table' | 'view' | 'procedure' | 'trigger';
```
eugene-mobile: **`ObjectType` belongs in `api.ts`, not in a UI component.**

This union type describes the domain objects that the `/api/table-ddl` endpoint understands — it's an API-layer concern. Defining it here creates an inverted dependency: `api.ts` has to stay in sync with `ShowSchemaDialog.tsx`, and any future consumer of `api.tableDdl` has to import from a component to get the type.

Move it to `api.ts`:
```ts
// api.ts
export type ObjectType = 'table' | 'view' | 'procedure' | 'trigger';
```
Then import it into `ShowSchemaDialog.tsx` and `SchemaBrowser.tsx` from `'../api'`.

src/api.ts
```diff
@@ -82,8 +82,9 @@ export const api = {
     return request<SchemaData>(`/api/schema?schema=${encodeURIComponent(schema)}`);
   },
 
-  tableDdl(schema: string, table: string) {
-    return request<{ ddl: string }>(`/api/table-ddl?schema=${encodeURIComponent(schema)}&table=${encodeURIComponent(table)}`);
+  tableDdl(schema: string, name: string, type?: string) {
```
eugene-mobile: **`type` param should use `ObjectType`, not `string`.**

Typing it as `string` means the compiler won't catch a caller passing `'veiw'` or `'Table'`. Once `ObjectType` is moved to `api.ts` (see comment on `ShowSchemaDialog.tsx`), tighten this:

```ts
tableDdl(schema: string, name: string, type?: ObjectType) {
```

src/components/SchemaBrowser.tsx
```diff
@@ -190,9 +190,35 @@ export function SchemaBrowser({ schema, activeTable, onTableSelect, onSchemaChan
               </div>
             ))}
 
-            {expanded[g.key] && g.key !== 'tables' && (
-              <div style={s.emptyGroup}>No {g.label.toLowerCase()}</div>
-            )}
+            {expanded[g.key] && g.key !== 'tables' && (() => {
```
eugene-mobile: **Same IIFE-in-JSX pattern flagged in PR #50.**

The `{condition && (() => { ... })()}` form is hard to scan. Since the logic here is non-trivial (derive type, derive items, filter, branch on empty), a small helper component or pre-computed variables before the main `return` would be cleaner:

```tsx
function NonTableGroupItems({ groupKey, schema, filter, t, onOpen, onContextMenu }: ...) {
  const objType: ObjectType = groupKey === 'views' ? 'view' : groupKey === 'procedures' ? 'procedure' : 'trigger';
  const items = (groupKey === 'views' ? schema.views : groupKey === 'procedures' ? schema.procedures : schema.triggers) ?? [];
  const filtered = filter ? items.filter(n => n.toLowerCase().includes(filter.toLowerCase())) : items;
  if (filtered.length === 0) return <div style={...}>No {groupKey}</div>;
  return <>{filtered.map(name => <div key={name} onClick={() => onOpen(name, objType)} ...>{name}</div>)}</>;
}
```

This also makes the parent JSX flat and easy to follow.

Issue #1, opened 2026-04-22
Ability to display table schema


eugene-mobile:
We should be able to display table schema. The most logical way to do so is right-click on the table and "Show Schema" button. Schema would be displayed in a modal window with the ability to copy it over.

PR #66, opened 2026-04-26
fix: pin USE and query to the same pool connection

eugene-mobile:
## Summary

- `pool.query()` borrows a connection, executes, and releases it immediately. Two back-to-back `pool.query()` calls can land on **different connections**, so `USE \`schema\`` had no guaranteed effect on the subsequent query.
- Fixed in `query.ts` by wrapping both calls in a single `pool.getConnection()` / `conn.release()` bracket (inner `try/finally` ensures the connection is always returned to the pool).
- Fixed in `mcp.ts` by replacing the `useSchema(pool, schema)` helper with a `withSchema<T>(pool, schema, fn)` wrapper; both `execute_query` and `execute_write` tool handlers now pass their logic as a callback so everything runs on one connection.

## Test plan

- [x] `npm test` in `server/` — 11 tests, all green
- [x] **Connection pinning** — asserts both `USE` and the query call `conn.query`, not `pool.query`, and that it is the *same* mock connection object
- [x] **Error paths** — connection is released even when `USE` throws or the query throws
- [x] **Input validation** — 400 on missing/blank sql
- [x] **Serialization** — Buffer→hex, Date→`YYYY-MM-DD HH:MM:SS`, BigInt→string, null, DML affectedRows

Closes issue #1 from the code analysis.

### DIFF
--- a/server/src/routes/query.test.ts
+++ b/server/src/routes/query.test.ts
@@ -1,0 +1,212 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import request from 'supertest';
+import express from 'express';
+
+vi.mock('../db.js', () => ({ getPool: vi.fn() }));
+
+import { getPool } from '../db.js';
+import { postQuery } from './query.js';
+
+// Minimal FieldPacket shape that postQuery reads from
+function makeField(name: string, flags = 0, columnType = 253) {
+  return { name, orgName: name, table: 't', orgTable: 't', flags, columnType, type: columnType };
+}
+
+function makeApp() {
+  const app = express();
+  app.use(express.json());
+  app.post('/api/query', postQuery);
+  return app;
+}
+
+describe('postQuery – connection pinning', () => {
+  beforeEach(() => vi.clearAllMocks());
+
+  it('issues USE and SELECT on the same connection when schema is provided', async () => {
+    const mockConn = {
+      query: vi.fn()
+        .mockResolvedValueOnce([[], []])                          // USE `mydb`
+        .mockResolvedValueOnce([                                  // SELECT * FROM users
+          [{ id: 1, name: 'Alice' }],
+          [makeField('id', /* PRI_KEY */ 2), makeField('name')],
+        ]),
+      release: vi.fn(),
+    };
+    vi.mocked(getPool).mockReturnValue({ getConnection: vi.fn().mockResolvedValue(mockConn) } as any);
+
+    const res = await request(makeApp())
+      .post('/api/query')
+      .send({ sql: 'SELECT * FROM users', schema: 'mydb' });
+
+    expect(res.status).toBe(200);
+
+    // Both calls landed on the same mock connection object — not on pool directly
+    expect(mockConn.query).toHaveBeenCalledTimes(2);
+    expect(mockConn.query).toHaveBeenNthCalledWith(1, 'USE `mydb`');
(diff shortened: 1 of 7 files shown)


server/src/routes/query.ts
```diff
@@ -12,74 +12,78 @@ export const postQuery: RequestHandler = async (req, res) => {
 
   try {
     const pool = getPool();
+    const conn = await pool.getConnection();
```
eugene-mobile: This inlines the same connection-pinning pattern (`getConnection` → USE → query → `release` in `try/finally`) that `mcp.ts` extracts into `withSchema<T>`. If the pattern ever needs to change (e.g. adding a timeout, logging the connection ID), there are now two places to update. Worth extracting `withSchema` into `db.ts` and importing it here in a follow-up.

server/src/routes/query.ts
```diff
@@ -12,74 +12,78 @@ export const postQuery: RequestHandler = async (req, res) => {
...
+      const serialized = rows.map(row => {
+        const out: Record<string, unknown> = {};
+        for (const col of columns) {
+          const val = row[col];
+          if (val === null || val === undefined) {
+            out[col] = null;
+          } else if (Buffer.isBuffer(val)) {
+            out[col] = val.toString('hex');
+          } else if (val instanceof Date) {
+            out[col] = val.toISOString().replace('T', ' ').replace(/\.\d{3}Z$/, '');
```
eugene-mobile, line 69: `.toISOString()` always outputs UTC regardless of the Node process timezone or the MySQL server timezone. MySQL `DATETIME` has no timezone — mysql2 gives you a `Date` object whose value depends on how the driver interprets the server's local time. If either side is non-UTC the displayed hour will be wrong. Pre-existing issue, but since you're adding tests that pin this behavior it's a good moment to either add a comment (`// assumes server and DB are both UTC`) or make the intent explicit: `val.toLocaleString('sv', { timeZone: 'UTC' }).replace('T', ' ')`.

server/src/routes/query.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    vi.mocked(getPool).mockReturnValue({ getConnection: vi.fn().mockResolvedValue(mockConn) } as any);
+
+    const res = await request(makeApp())
+      .post('/api/query')
+      .send({ sql: "UPDATE users SET active = 0 WHERE role = 'guest'" });
+
+    expect(res.status).toBe(200);
+    expect(res.body.columns).toEqual([]);
+    expect(res.body.rows).toEqual([]);
+    expect(res.body.affectedRows).toBe(3);
```
eugene-mobile, line 226: Every other test case asserts `expect(mockConn.release).toHaveBeenCalledTimes(1)`. This one skips it — easy one-liner to add for consistency.

eugene-mobile (commented):
Fix is correct and well-tested. Three non-blocking notes below.

eugene-mobile:
All three review comments addressed in f008696:

**withSchema duplication** — extracted from mcp.ts into db.ts and exported. Both query.ts and mcp.ts now import it from the same place. The pool argument was dropped (withSchema calls getPool() internally), keeping the call sites clean. The i.mock factory in the test was updated to supply withSchema mirroring the real implementation so all connection-pinning assertions still hold.

**UTC assumption on Date serialisation** — added an explicit comment on the .toISOString() line in query.ts noting the UTC assumption and when it would be wrong.

**Missing 
elease assertion** — added expect(mockConn.release).toHaveBeenCalledTimes(1) to the DML test case.

PR #72, opened 2026-04-26
perf: load schema with Promise.all instead of sequential awaits

### LINKED ISSUE #67: perf: schema browser load makes 5 sequential DB round-trips — use Promise.all()
## Problem

`server/src/routes/schema.ts` — `getSchema` handler fires five `await pool.query(...)` calls back-to-back:

1. `information_schema.TABLES` (table list)
2. `information_schema.COLUMNS` (all columns for schema)
3. `information_schema.VIEWS`
4. `information_schema.ROUTINES` (procedures)
5. `information_schema.TRIGGERS`

Each call waits for the previous one to complete before starting, so total latency is the sum of all five round-trips. On a remote database with even 20 ms per query that is 100 ms of avoidable serial waiting every time a schema is loaded or refreshed.

All five queries are fully independent — none of their results are used as input to another.

## Expected fix

Replace the sequential awaits with `Promise.all`:

```ts
const [
  [tables],
  [columns],
  [views],
  [procedures],
  [triggers],
] = await Promise.all([
  pool.query<RowDataPacket[]>(`SELECT ... FROM information_schema.TABLES ...`, [schema]),
  pool.query<RowDataPacket[]>(`SELECT ... FROM information_schema.COLUMNS ...`, [schema]),
  pool.query<RowDataPacket[]>(`SELECT ... FROM information_schema.VIEWS ...`, [schema]),
  pool.query<RowDataPacket[]>(`SELECT ... FROM information_schema.ROUTINES ...`, [schema]),
  pool.query<RowDataPacket[]>(`SELECT ... FROM information_schema.TRIGGERS ...`, [schema]),
]);
```

This reduces schema-load latency to the slowest single query rather than their sum.

## File

`server/src/routes/schema.ts` — `getSchema` handler (lines 30–69)

eugene-mobile:
## Summary

- `getSchema` in `server/src/routes/schema.ts` was issuing 5 `await pool.query()` calls back-to-back: tables → columns → views → procedures → triggers
- Each call blocked the next, making total latency the **sum** of all five round-trips instead of the slowest one
- All 5 queries are completely independent (none feeds into another), so they are safe to parallelise
- Replace with a single `Promise.all([...])` — the destructuring keeps variable names identical so the response-assembly logic below is untouched

## Test plan

- [x] App connects and schema browser loads correctly (4 tables, 1 view, 1 procedure, 1 trigger all present)
- [x] Table click → SELECT query runs and returns results
- [x] View DDL modal opens and shows correct `CREATE VIEW` statement
- [x] `/api/schema?schema=demo` responds in ~10 ms on localhost
- [x] TypeScript build passes (`npm run build`)

Closes #67

### DIFF
--- a/server/src/routes/schema.ts
+++ b/server/src/routes/schema.ts
@@ -27,46 +27,50 @@ export const getSchema: RequestHandler = async (req, res) => {
   try {
     const pool = getPool();
 
-    const [tables] = await pool.query<mysql.RowDataPacket[]>(
-      `SELECT TABLE_NAME AS name, TABLE_ROWS AS row_count, TABLE_COMMENT AS comment
-       FROM information_schema.TABLES
-       WHERE TABLE_SCHEMA = ? AND TABLE_TYPE = 'BASE TABLE'
-       ORDER BY TABLE_NAME`,
-      [schema]
-    );
-
-    const [columns] = await pool.query<mysql.RowDataPacket[]>(
-      `SELECT TABLE_NAME AS tbl, COLUMN_NAME AS col, COLUMN_TYPE AS col_type,
-              DATA_TYPE AS data_type,
-              IF(COLUMN_KEY = 'PRI', 1, 0) AS is_pk,
-              IF(IS_NULLABLE = 'YES', 1, 0) AS nullable,
-              COLUMN_DEFAULT AS col_default,
-              EXTRA AS extra,
-              COLUMN_COMMENT AS comment
-       FROM information_schema.COLUMNS
-       WHERE TABLE_SCHEMA = ?
-       ORDER BY TABLE_NAME, ORDINAL_POSITION`,
-      [schema]
-    );
-
-    const [views] = await pool.query<mysql.RowDataPacket[]>(
-      `SELECT TABLE_NAME AS name FROM information_schema.VIEWS
-       WHERE TABLE_SCHEMA = ? ORDER BY TABLE_NAME`,
-      [schema]
-    );
-
-    const [procedures] = await pool.query<mysql.RowDataPacket[]>(
-      `SELECT ROUTINE_NAME AS name FROM information_schema.ROUTINES
-       WHERE ROUTINE_SCHEMA = ? AND ROUTINE_TYPE = 'PROCEDURE'
-       ORDER BY ROUTINE_NAME`,
-      [schema]
-    );
-
-    const [triggers] = await pool.query<mysql.RowDataPacket[]>(
-      `SELECT TRIGGER_NAME AS name FROM information_schema.TRIGGERS
-       WHERE TRIGGER_SCHEMA = ? ORDER BY TRIGGER_NAME`,
-      [schema]
-    );
+    const [
+      [tables],
+      [columns],
+      [views],
+      [procedures],
+      [triggers],
+    ] = await Promise.all([
+      pool.query<mysql.RowDataPacket[]>(
+        `SELECT TABLE_NAME AS name, TABLE_ROWS AS row_count, TABLE_COMMENT AS comment
+         FROM information_schema.TABLES
+         WHERE TABLE_SCHEMA = ? AND TABLE_TYPE = 'BASE TABLE'
+         ORDER BY TABLE_NAME`,
+        [schema],
+      ),
+      pool.query<mysql.RowDataPacket[]>(
+        `SELECT TABLE_NAME AS tbl, COLUMN_NAME AS col, COLUMN_TYPE AS col_type,
+                DATA_TYPE AS data_type,
+                IF(COLUMN_KEY = 'PRI', 1, 0) AS is_pk,
+                IF(IS_NULLABLE = 'YES', 1, 0) AS nullable,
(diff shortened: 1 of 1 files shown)


server/src/routes/schema.ts
```diff
@@ -27,46 +27,50 @@ export const getSchema: RequestHandler = async (req, res) => {
...
-       ORDER BY ROUTINE_NAME`,
-      [schema]
-    );
-
-    const [triggers] = await pool.query<mysql.RowDataPacket[]>(
-      `SELECT TRIGGER_NAME AS name FROM information_schema.TRIGGERS
-       WHERE TRIGGER_SCHEMA = ? ORDER BY TRIGGER_NAME`,
-      [schema]
-    );
+    const [
```
eugene-mobile: **[Blocking] Pool exhaustion under concurrent requests**

`Promise.all` fires all 5 queries simultaneously. The pool is configured with `connectionLimit: 5` (see `db.ts` line 31). A single call to `getSchema` now needs up to 5 connections at the same instant. If even one other request is already active (e.g. the user running a query in the editor), `Promise.all` will stall waiting for a free connection slot — and total latency can exceed the old sequential approach under moderate load.

Options:
1. Raise `connectionLimit` to ≥ 10 and document the reason in `db.ts`.
2. Acquire one connection with `pool.getConnection()`, run all 5 queries sequentially on it inside `try/finally { conn.release() }`. The queries are independent of each other, so correctness is unchanged, and only one pool slot is consumed per schema load.
3. If accepting the current behaviour, add a comment here documenting the pool-size dependency so the next person who tunes `connectionLimit` understands the risk.

As written, two concurrent schema loads can fully exhaust the pool.

server/src/routes/schema.ts
```diff
@@ -27,46 +27,50 @@ export const getSchema: RequestHandler = async (req, res) => {
...
+         WHERE ROUTINE_SCHEMA = ? AND ROUTINE_TYPE = 'PROCEDURE'
+         ORDER BY ROUTINE_NAME`,
+        [schema],
+      ),
+      pool.query<mysql.RowDataPacket[]>(
+        `SELECT TRIGGER_NAME AS name FROM information_schema.TRIGGERS
+         WHERE TRIGGER_SCHEMA = ? ORDER BY TRIGGER_NAME`,
+        [schema],
+      ),
+    ]);
```
eugene-mobile, line 70: **[Non-blocking] `Promise.all` fail-fast hides which query caused the error**

If one of the 5 queries rejects (e.g. insufficient privileges on `information_schema.TRIGGERS`), the `catch` block receives the raw error with no indication of which query failed. Debugging requires trial-and-error.

Consider tagging errors with the query name, or at minimum add a comment here noting the fail-fast behaviour so operators know to check MySQL grants for each `information_schema` table independently when this endpoint returns 500.

eugene-mobile (commented):
The parallelisation is correct and the PR goal is sound, but there are issues to address before merge.

**[Blocking — pre-existing, not introduced by this PR] `query.ts` line 17: Connection-pinning bug — `USE` and query run on different pool connections**

```ts
await pool.query(`USE \`${schema.replace(/`/g, '')}\``);
// ...
const [rows, fields] = await pool.query(sql);
```

Both calls go through `pool.query()`, which checks out *any available connection* from the pool each time. Under concurrent requests the `USE` and the subsequent `pool.query(sql)` can land on **different** connections, so the `USE` has no effect on the query that follows it. This will cause intermittent wrong-database errors under load. The fix is to acquire one connection and use it for both statements:

```ts
const conn = await pool.getConnection();
try {
  if (schema) await conn.query(`USE \`${schema.replace(/`/g, '')}\``);
  const [rows, fields] = await conn.query(sql) as [RowDataPacket[], FieldPacket[]];
  // ...
} finally {
  conn.release();
}
```

**[Non-blocking] `schema.ts` line 21: Whitespace-only schema name passes the guard**

`if (!schema)` is falsy only for `''`, `null`, and `undefined`. A request like `?schema=%20` passes the guard and reaches the parameterized queries, which return empty result sets silently instead of a 400 error. Change to match `tableDdl.ts` line 12:

```ts
if (!schema || !schema.trim()) {
  res.status(400).json({ error: 'schema query param is required.' });
  return;
}
```

eugene-mobile:
All four review points addressed in d17a7ac:

**[Blocking] Pool exhaustion** — replaced pool.query() x5 in Promise.all with getConnection() + conn.query() x5 on that one connection + 
elease() in inally. mysql2 serialises concurrent queries on a single connection, so correctness is unchanged and only 1 pool slot is consumed per schema load instead of 5.

**[Blocking] Connection-pinning in query.ts** — fixed the pre-existing bug: both USE \schema\` and the user query now run on the same connection acquired via getConnection(), released in inally.

**[Non-blocking] Whitespace schema guard** — changed if (!schema) to if (!schema?.trim()) so ?schema=%20 returns 400.

**[Non-blocking] Fail-fast error visibility** — added a comment above the Promise.all block noting the fail-fast behaviour and that per-table information_schema grants should be checked independently when the endpoint returns 500.

PR #74, opened 2026-04-27
fix: remove duplicate loadSchema calls on connect and schema switch

### LINKED ISSUE #69: bug: schema fetched twice on initial connection (double loadSchema call)
## Problem

When a connection is established, `handleConnect` in `src/App.tsx` calls `loadSchema` explicitly:

```ts
// line 80
if (initial) await loadSchema(initial);
// ...
setActiveSchema(initial);   // line 87 — triggers the useEffect below
```

Then a `useEffect` watches `activeSchema` and fires `loadSchema` whenever it changes:

```ts
// lines 293-295
useEffect(() => {
  if (connected && activeSchema) loadSchema(activeSchema);
}, [activeSchema, connected, loadSchema]);
```

Because `setActiveSchema(initial)` is called immediately after the explicit `loadSchema` call, and `connected` is set to `true` in the same batch, the effect fires and calls `loadSchema` a second time. The result is two identical `GET /api/schema?schema=<name>` requests sent to the server on every new connection.

## Steps to reproduce

1. Open the Network tab in browser DevTools.
2. Connect to a MySQL server with any schema.
3. Observe two identical requests to `/api/schema?schema=<dbname>` fired in rapid succession.

## Expected behaviour

The schema should be fetched exactly once on connection. Either:

- Remove the explicit `loadSchema(initial)` call from `handleConnect` and let the `useEffect` handle it (simplest fix), or
- Remove the `useEffect` entirely and call `loadSchema` explicitly wherever the schema needs to reload (more predictable data-flow).

## File

`src/App.tsx` — `handleConnect` (line 80) and the `useEffect` at lines 293-295.

eugene-mobile:
## Summary

- `handleConnect` called `loadSchema(initial)` explicitly, then set `activeSchema` and `connected` which fired the `useEffect` watcher — doubling every schema fetch on connect
- `handleSchemaChange` called `loadSchema(schema)` explicitly, then set `activeSchema` which again fired the `useEffect` — doubling every schema switch fetch
- Fix: remove both explicit calls and let the `useEffect` be the single authority for schema fetches

## Test plan

- [x] Opened DevTools Network tab, connected to MySQL — confirmed `GET /api/schema` fires exactly once (was twice before)
- [x] Schema browser populated correctly after connect
- [x] `npm run build` clean

Closes #69

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -64,6 +64,7 @@ export default function App() {
   }, []);
 
   const handleSchemaChange = useCallback((schema: string) => {
+    // Schema fetch is driven by the useEffect on activeSchema; only update state here.
     setActiveSchema(schema);
   }, []);
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -65,8 +65,7 @@ export default function App() {
 
   const handleSchemaChange = useCallback((schema: string) => {
     setActiveSchema(schema);
-    loadSchema(schema);
-  }, [loadSchema]);
+  }, []);
 
   const handleConnect = async (form: ConnectionForm) => {
     setIsConnecting(true);
@@ -77,7 +76,6 @@ export default function App() {
       const initial = form.database && list.includes(form.database)
         ? form.database
         : list[0] ?? '';
-      if (initial) await loadSchema(initial);
 
       const friendly = form.name.trim();
       setConnectionName(friendly || res.connectionName);


src/App.tsx
```diff
@@ -65,8 +65,7 @@ export default function App() {
 
   const handleSchemaChange = useCallback((schema: string) => {
     setActiveSchema(schema);
-    loadSchema(schema);
-  }, [loadSchema]);
+  }, []);
```
eugene-mobile, line 69: **Non-blocking: the empty-schema edge case is handled correctly but silently**

When the server returns an empty schema list, `initial` is `''`, `setActiveSchema('')` fires (line 86 in the new file), and the `useEffect` guard `if (connected && activeSchema)` at line 292 correctly suppresses the `/api/schema` fetch — no double call, no spurious request. This is the right behaviour.

The PR description and test plan only verify the happy path (schemas present, exactly one network call). Worth adding either a short inline comment at the `useEffect` guard or a test for `handleConnect` with an empty schema list, so a future reader does not reintroduce `if (initial) await loadSchema(initial)` thinking the guard was accidentally omitted.

eugene-mobile, line 69: **`handleSchemaChange` dep array is now `[]` — correct, but worth a note**

Removing `loadSchema` from the dependency array is safe here because `setActiveSchema` (a stable dispatch from `useState`) is the only thing called, and the `useEffect` at line 291–293 is the exclusive trigger for schema fetches. However, the empty `[]` looks like a lint suppression to unfamiliar readers. A one-line comment — `// schema fetch is driven by the useEffect below; only state update here` — would make the intent explicit without changing behaviour.

eugene-mobile (commented):
The duplicate-fetch fix is correct and the `useEffect`-as-single-authority pattern is the right approach. The two issues below are pre-existing (not introduced by this PR) but surfaced during review and should be addressed.

---

**Pre-existing bug in `server/src/routes/schema.ts` line 31: `getConnection()` is outside the outer try/catch**

`getPool().getConnection()` in `getSchema` sits *before* the `try` block (line 32). If the pool is exhausted or the DB is unreachable, the error escapes the `catch` at line 109, Express's default handler returns a plain-text HTML 500, and the client cannot parse it as JSON. The correct pattern — which `withSchema` in `db.ts` already uses — is to wrap `getConnection()` inside the outer try. A `withConnection` helper analogous to `withSchema` would enforce this in one place.

**Pre-existing misleading comment in `server/src/routes/schema.ts` line 33: `Promise.all` on one connection has no parallelism**

mysql2 serialises queries on a single `PoolConnection` at the protocol level. `Promise.all` here queues all five queries sequentially — it does not execute them in parallel. The surrounding comment justifies using one connection (correct), but a future maintainer may read `Promise.all` as evidence of genuine concurrency and try to split the queries across separate connections expecting a speedup. Suggest either sequential `await` calls or an explicit note: `// Promise.all here is for concise destructuring only — mysql2 serialises all queries on one connection`.

**Missing tests: `getSchema` has no test file**

The `query.test.ts` suite covers `getConnection` failure for `postQuery` (via the `withSchema` mock), but there are no tests for `getSchema` at all. Given the `getConnection` bug above, the first priority would be a test confirming a JSON error body is returned when `getConnection` rejects.

src/App.tsx
eugene-mobile: Good call — pushed 56638f6 with a one-line note above the setter. Empty `[]` next to a state setter is exactly the kind of thing that reads like lint suppression without context, so this is a fair WHY-is-non-obvious case for an inline comment.

eugene-mobile: Going to leave this one as-is. The `if (connected && activeSchema)` guard is self-documenting at the call site — `activeSchema` truthiness is plain reading, not a subtle invariant. The repo convention (.claude/agents/software-engineer.md) is to add comments only when the WHY is non-obvious, and the regression risk is already covered: the Common Pitfalls table in that same file flags "Schema loaded twice on connect" with the exact misstep to avoid. Adding a frontend test just for the empty-list path is also a meaningful scope expansion (no frontend test harness on this repo yet) and better tracked as its own issue if we want to stand one up.

PR #76, opened 2026-04-27
Electron app

eugene-mobile:
Create local distributable


electron/main.cjs
```diff
@@ -0,0 +1,135 @@
...
+function waitForServer() {
+  return new Promise((resolve) => {
+    const attempt = () => {
+      http
+        .get(`${APP_URL}/api/connect/status`, resolve)
+        .on('error', () => setTimeout(attempt, 300));
+    };
+    attempt();
+  });
+}
```
eugene-mobile, line 51: **Blocking: `waitForServer()` never rejects — app hangs forever if the server fails to start.**

If `utilityProcess.fork` fails silently, or if port 3001 is already in use, `serverProc.stderr` logs the error but `waitForServer` just keeps retrying at 300 ms intervals with no upper bound. The app will appear frozen at startup with no feedback to the user.

Add a retry cap:

```js
function waitForServer(maxAttempts = 50) {
  return new Promise((resolve, reject) => {
    let attempts = 0;
    const attempt = () => {
      if (attempts++ >= maxAttempts) {
        reject(new Error('Server did not start in time'));
        return;
      }
      http
        .get(`${APP_URL}/api/connect/status`, resolve)
        .on('error', () => setTimeout(attempt, 300));
    };
    attempt();
  });
}
```

And handle the rejection in `whenReady` — e.g. `dialog.showErrorBox(...)` so the user sees a message instead of a blank frozen window.

server/src/index.ts
```diff
@@ -43,6 +44,13 @@ app.post('/mcp', mcpHandler);
 app.get('/mcp', mcpHandler);
 app.delete('/mcp', mcpHandler);
 
+// Serve the Vite frontend in Electron production builds
+const staticPath = process.env['STATIC_PATH'];
+if (staticPath) {
+  app.use(express.static(staticPath));
+  app.get('*', (_req, res) => res.sendFile(join(staticPath, 'index.html')));
```
eugene-mobile, line 52: **Non-blocking: catch-all shadows any API routes registered after this block.**

`app.get('*', ...)` is correctly placed after all current `/api/*` routes, so today it's fine. But any future route added below this block will silently return `index.html` with a 200 instead of a proper response. A short comment prevents that footgun:

```ts
// SPA fallback — must stay LAST; any /api route added after this will be shadowed
app.get('*', (_req, res) => res.sendFile(join(staticPath, 'index.html')));
```

electron/main.cjs
```diff
@@ -0,0 +1,135 @@
...
+    webPreferences: { nodeIntegration: false, contextIsolation: true },
+    titleBarStyle: process.platform === 'darwin' ? 'hiddenInset' : 'default',
+  });
+
+  const url = isDev ? VITE_URL : APP_URL;
+  win.loadURL(url);
+
+  // Retry if Vite/Express dev servers haven't started yet
+  win.webContents.on('did-fail-load', (_e, code) => {
+    if (isDev && code === -102) setTimeout(() => win?.loadURL(url), 1000);
```
eugene-mobile: **Non-blocking: magic number `-102` should be a named constant.**

`-102` is Chromium's `ERR_CONNECTION_REFUSED`. Define it at the top of the file:

```js
const ERR_CONNECTION_REFUSED = -102;
```

and use `code === ERR_CONNECTION_REFUSED` in the handler.

eugene-mobile (commented):
Solid Electron integration — structure is clean and the boolean type propagation is complete. One blocking issue around server startup failure handling, plus two minor non-blocking notes.

PR #77, opened 2026-04-27
fix: connection modal can be dismissed without connecting

### LINKED ISSUE #52: Connection modal cannot be dismissed - no Cancel button, Escape, or click-outside
## Problem
When the connection modal appears (on first load or after a page refresh), there is no way to dismiss it:
- No **Cancel** or **Close (x)** button
- **Escape** key does not close the modal
- Clicking outside the modal (backdrop) does not close it

Users are fully trapped in the modal until they click **Connect**.

## Steps to reproduce
1. Refresh the page while connected to a database
2. The connection modal appears pre-filled with the saved connection
3. Press Escape - nothing happens
4. Click outside the modal - nothing happens
5. There is no Cancel or X button visible

## Expected behaviour
- Escape key dismisses the modal (returning to main UI if already connected)
- Clicking the backdrop dismisses it
- A visible Cancel button is present

## Notes
Inconsistent with the Insert Row modal, which correctly dismisses on Escape.

eugene-mobile:
Fixes #52

## Summary
- **Cancel button** added to the connection modal footer; hidden while a connection attempt is in progress
- **Escape key** dismisses the modal (unless a connection is in progress)
- **Backdrop click** dismisses the modal (clicking outside the modal panel)
- **Disconnected status pill** in the TopBar is now clickable to reopen the modal after dismissing

## How it works
Introduces a `showConnectionModal` boolean in `App.tsx` (independent of the `connected` flag) so the modal can be hidden without changing connection state. On disconnect the modal reappears automatically; on successful connect it hides. The `onDismiss` prop is optional on `ConnectionManager` so cancel affordances are only rendered when a dismiss handler is wired up.

## Test plan
- [ ] Open app fresh — connection modal appears; Cancel, Escape, and backdrop click all dismiss it
- [ ] After dismissing, the status pill shows "Not connected" and is clickable — clicking reopens the modal
- [ ] Successful connect hides the modal; Cancel is not visible while the Connecting spinner is active
- [ ] Disconnect via the TopBar menu — modal reappears automatically
- [ ] Escape on the connection name autocomplete dropdown still only closes the dropdown (not the modal)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -335,6 +335,9 @@ export default function App() {
   const bodyStyle: CSSProperties = { display: 'flex', flex: 1, overflow: 'hidden' };
   const mainStyle: CSSProperties = { display: 'flex', flexDirection: 'column', flex: 1, overflow: 'hidden' };
 
+  const handleDismissModal = useCallback(() => setShowConnectionModal(false), []);
+  const handleOpenConnection = useCallback(() => setShowConnectionModal(true), []);
+
   return (
     <div style={appStyle}>
       <TopBar
@@ -347,7 +350,7 @@ export default function App() {
         connectionHost={connectionHost}
         connectionStatus={connected ? 'connected' : 'disconnected'}
         onDisconnect={handleDisconnect}
-        onOpenConnection={connected ? undefined : () => setShowConnectionModal(true)}
+        onOpenConnection={connected ? undefined : handleOpenConnection}
         mcpWritesAllowed={mcpWritesAllowed}
         mcpUrl={mcpUrl}
         onToggleMcpWrites={handleToggleMcpWrites}
@@ -410,7 +413,7 @@ export default function App() {
           onConnect={handleConnect}
           isConnecting={isConnecting}
           error={connectionError}
-          onDismiss={() => setShowConnectionModal(false)}
+          onDismiss={handleDismissModal}
           t={t}
         />
       )}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -44,6 +44,7 @@ export default function App() {
   }, [themeName]);
 
   const [connected, setConnected] = useState(false);
+  const [showConnectionModal, setShowConnectionModal] = useState(true);
   const [isConnecting, setIsConnecting] = useState(false);
   const [connectionError, setConnectionError] = useState<string | null>(null);
   const [connectionName, setConnectionName] = useState('Not connected');
@@ -102,6 +103,7 @@ export default function App() {
       setSchemas(list);
       setActiveSchema(initial);
       setConnected(true);
+      setShowConnectionModal(false);
 
       // Persist non-secret fields so the user doesn't retype them next time.
       if (friendly) {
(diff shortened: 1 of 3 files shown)


src/components/ConnectionManager.tsx
```diff
@@ -39,6 +40,13 @@ export function ConnectionManager({ onConnect, isConnecting, error, t }: Connect
   const [testing, setTesting] = useState(false);
   const [testResult, setTestResult] = useState<{ ok: true } | { ok: false; error: string } | null>(null);
 
+  useEffect(() => {
+    if (!onDismiss || isConnecting) return;
+    const key = (e: KeyboardEvent) => { if (e.key === 'Escape') onDismiss(); };
+    window.addEventListener('keydown', key);
+    return () => window.removeEventListener('keydown', key);
+  }, [onDismiss, isConnecting]);
```
eugene-mobile, line 48: **Blocking bug — Escape closes the modal when it should only close the autocomplete dropdown.**

This `useEffect` registers a `window` keydown listener. The name-input's `onKeyDown` handler (line ~161) also handles Escape when the autocomplete dropdown is open, calling `e.preventDefault()` and `setSuggestOpen(false)`. However, `e.preventDefault()` and React's `e.stopPropagation()` do **not** stop the native DOM event from reaching `window` listeners — only `e.nativeEvent.stopImmediatePropagation()` does that. As a result, pressing Escape to close the dropdown will simultaneously call `onDismiss()` and dismiss the entire modal.

This directly contradicts the PR's own test-plan item: *"Escape on the connection name autocomplete dropdown still only closes the dropdown (not the modal)."*

Fix: add `e.nativeEvent.stopImmediatePropagation()` in the input's `onKeyDown` Escape branch:
```tsx
else if (e.key === 'Escape') {
  e.preventDefault();
  e.nativeEvent.stopImmediatePropagation(); // prevent the modal-level window listener from also firing
  setSuggestOpen(false);
}
```

src/App.tsx
```diff
@@ -401,11 +405,12 @@ export default function App() {
         </div>
       </div>
 
-      {!connected && (
+      {!connected && showConnectionModal && (
         <ConnectionManager
           onConnect={handleConnect}
           isConnecting={isConnecting}
           error={connectionError}
+          onDismiss={() => setShowConnectionModal(false)}
```
eugene-mobile: **Non-blocking — `onDismiss` (and `onOpenConnection` on line 350) are new function references on every render, causing the Escape listener in `ConnectionManager` to churn.**

Because `() => setShowConnectionModal(false)` is an inline arrow function, `onDismiss` fails referential equality on every re-render. The `useEffect` in `ConnectionManager` has `onDismiss` as a dependency, so it removes and re-adds the `window` keydown listener on every render. There is no visible breakage today, but it wastes work and is fragile if anything causes frequent re-renders during a connection attempt.

Fix: memoize both callbacks:
```tsx
const handleDismissConnection = useCallback(() => setShowConnectionModal(false), []);
const handleOpenConnection    = useCallback(() => setShowConnectionModal(true),  []);
```
Then substitute them for the two inline arrows (here and on line 350).

eugene-mobile (commented):
One blocking bug (Escape on the autocomplete dropdown also dismisses the modal) and one non-blocking callback-stability issue; everything else looks correct.

PR #79, opened 2026-04-28
feat: integration tests against real MySQL via Docker Compose

### LINKED ISSUE #3: Integration testing for MySQL queries
We need integration testing with MySQL. MySQL should be running as a docker container.

eugene-mobile:
Closes #3

## Summary
- `docker-compose.yml` — MySQL 8.0 service on port 3307 (avoids collision with a local MySQL on 3306)
- `server/vitest.integration.config.ts` — separate Vitest config targeting `*.integration.test.ts`, sequential single-fork execution so tests share one DB connection cleanly
- `server/src/test-setup/global-setup.ts` — polls until MySQL is reachable, creates `helix_test` schema with `users` and `orders` tables, drops the schema on teardown
- `server/src/routes/crud.integration.test.ts` — 18 tests across `postQuery`, `postInsertRow`, `postUpdateCell`, `postDeleteRow` against a live MySQL instance
- `server/package.json` — adds `test:integration` script

## Usage

```bash
docker compose up -d
cd server && npm run test:integration
docker compose down
```

Unit tests (`npm test`) are unchanged and still run without Docker.

## What the integration tests cover
- `postQuery`: empty table, inserted rows, schema switching via USE, NULL serialisation, UPDATE affectedRows, bad SQL → 400, unknown schema → 400
- `postInsertRow`: happy path with insertId, NULL value, nonexistent table → 400
- `postUpdateCell`: update by PK, update to NULL, WHERE matches nothing → affectedRows 0
- `postDeleteRow`: delete by PK, WHERE matches nothing → affectedRows 0, compound WHERE

## Not included (follow-up)
- GitHub Actions CI workflow (can be added once this is merged)
- `schema` and `connect` route coverage


server/src/test-setup/global-setup.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+async function waitForMySQL(): Promise<mysql.Connection> {
+  const deadline = Date.now() + TIMEOUT_MS;
+  let lastError: unknown;
+  while (Date.now() < deadline) {
+    try {
+      const conn = await mysql.createConnection(config);
+      await conn.ping();
+      return conn;
```
eugene-mobile: **Blocking — connection leak on ping failure.**

If `createConnection()` succeeds but `conn.ping()` throws, the connection is leaked and never closed. Over up to 60 retries (30 s / 0.5 s) this can accumulate many open connections.

Fix — split the try/catch so a failed ping still closes the connection:
```ts
const conn = await mysql.createConnection(config);
try {
  await conn.ping();
  return conn;
} catch (pingErr) {
  await conn.end();
  throw pingErr; // re-throw so outer catch retries
}
```

server/src/test-setup/global-setup.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    CREATE TABLE orders (
+      id         INT AUTO_INCREMENT PRIMARY KEY,
+      user_id    INT NOT NULL,
+      amount     DECIMAL(10,2) NOT NULL,
+      created_at DATETIME DEFAULT CURRENT_TIMESTAMP
+    );
+  `);
+
+  await conn.end();
```
eugene-mobile: **Blocking — connection not closed if schema creation throws.**

`conn.end()` is only called on the happy path. If the multi-statement DDL query fails partway through, the connection leaks. Wrap in `try/finally`:
```ts
const conn = await waitForMySQL();
try {
  await conn.query(`...`);
} finally {
  await conn.end();
}
```
The same pattern is needed in `teardown()` at line 59.

eugene-mobile (commented):
Solid integration test suite overall — real DB, sequential execution, proper teardown. Three issues worth fixing before merge: two connection leaks in global-setup and a misleading comment in the Vitest config.

server/src/test-setup/global-setup.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export async function teardown() {
+  const conn = await mysql.createConnection(config);
+  await conn.query('DROP DATABASE IF EXISTS helix_test');
+  await conn.end();
```
eugene-mobile: **Blocking — connection not closed if DROP DATABASE throws.**

Same issue as `setup()`: `conn.end()` is not in a `finally` block.
```ts
const conn = await mysql.createConnection(config);
try {
  await conn.query('DROP DATABASE IF EXISTS helix_test');
} finally {
  await conn.end();
}
```

server/vitest.integration.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    environment: 'node',
+    include: ['src/**/*.integration.test.ts'],
+    globalSetup: './src/test-setup/global-setup.ts',
+    testTimeout: 15000,
+    hookTimeout: 15000,
+    // Run integration tests sequentially — each test shares the same DB connection
```
eugene-mobile: **Non-blocking — misleading comment.**

"each test shares the same DB connection" is inaccurate — tests call the app over HTTP and each request goes through the pool independently. The real reason for `singleFork: true` is to run all tests in a single process so they share the module-level `raw` and `app` state and execute sequentially (preventing concurrent writes from racing). Suggest:
```ts
// Run all tests in one process to share module-level state (app, raw conn)
// and prevent concurrent writes from racing between tests.
```

PR #86, opened 2026-04-28
Store DB passwords via OS keychain in Electron mode

eugene-mobile:
## Summary
- Optional **Save password** checkbox in the connection form. When checked, the password is persisted using Electron's \`safeStorage\` (Keychain on macOS, DPAPI on Windows, libsecret on Linux).
- Checkbox only appears when running inside Electron *and* the OS reports encryption as available — browser / \`npm run dev\` mode is unchanged.
- Picking a saved connection auto-fills the stored password if one is on file. **Forget** also wipes the stored password.

## Implementation
- \`electron/preload.cjs\` (new) exposes \`window.electronAPI.passwords\` (\`available\`, \`save\`, \`load\`, \`delete\`) via \`contextBridge\` over IPC.
- \`electron/main.cjs\` registers IPC handlers, encrypts via \`safeStorage\`, and persists base64 ciphertext to \`passwords.json\` under \`app.getPath('userData')\` with \`mode 0600\`.
- \`SavedConnection\` gains a \`savePassword\` flag (in localStorage) so the form knows to load the password on apply.
- \`ConnectionManager\` adds the checkbox and password preload effect; \`App.tsx\` saves/deletes the password on successful connect.

## Test plan
- [ ] In packaged or \`npm run electron:dev\` build: connect with **Save password** checked → restart app → connection auto-populates the password field.
- [ ] Uncheck **Save password** and reconnect → stored password is removed.
- [ ] **Forget** a saved connection → both the localStorage entry and the keychain entry are deleted.
- [ ] Run \`npm run dev\` (browser) → checkbox is hidden, no regressions.
- [ ] On Linux without a keychain available → checkbox is hidden (\`isEncryptionAvailable\` returns false).

### DIFF
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -1,8 +1,9 @@
 'use strict';
 
-const { app, BrowserWindow, Tray, Menu, nativeImage, utilityProcess, dialog } = require('electron');
+const { app, BrowserWindow, Tray, Menu, nativeImage, utilityProcess, dialog, ipcMain, safeStorage } = require('electron');
 const path = require('path');
 const http = require('http');
+const fs = require('fs');
 
 const isDev = !app.isPackaged;
 const PORT = 3001;
@@ -52,6 +53,65 @@ function waitForServer(maxAttempts = 50) {
   });
 }
 
+// ── Saved passwords (encrypted via OS keychain) ───────────────────────────────
+
+function passwordsFile() {
+  return path.join(app.getPath('userData'), 'passwords.json');
+}
+
+let passwordsCache = null;
+function readPasswords() {
+  if (passwordsCache) return passwordsCache;
+  try {
+    passwordsCache = JSON.parse(fs.readFileSync(passwordsFile(), 'utf8'));
+  } catch {
+    passwordsCache = {};
+  }
+  return passwordsCache;
+}
+function writePasswords(obj) {
+  passwordsCache = obj;
+  try {
+    fs.writeFileSync(passwordsFile(), JSON.stringify(obj, null, 2), { mode: 0o600 });
+  } catch (err) {
+    console.error('[passwords] write failed:', err);
+  }
+}
+
+ipcMain.handle('passwords:available', () => {
+  try { return safeStorage.isEncryptionAvailable(); } catch { return false; }
+});
+
+ipcMain.handle('passwords:save', (_e, name, password) => {
+  if (typeof name !== 'string' || !name || typeof password !== 'string') return;
+  if (!safeStorage.isEncryptionAvailable()) throw new Error('Encryption not available on this system');
+  const all = readPasswords();
+  all[name] = safeStorage.encryptString(password).toString('base64');
+  writePasswords(all);
+});
+
+ipcMain.handle('passwords:load', (_e, name) => {
+  if (typeof name !== 'string' || !name) return null;
+  const all = readPasswords();
+  const b64 = all[name];
+  if (!b64) return null;
+  try {
+    return safeStorage.decryptString(Buffer.from(b64, 'base64'));
+  } catch (err) {
+    console.error('[passwords] decrypt failed for', name, err);
+    return null;
+  }
+});
+
+ipcMain.handle('passwords:delete', (_e, name) => {
+  if (typeof name !== 'string' || !name) return;
+  const all = readPasswords();
+  if (name in all) {
+    delete all[name];
+    writePasswords(all);
+  }
+});
+
 // ── Icons ─────────────────────────────────────────────────────────────────────
 
 function loadIcon(filename) {
@@ -70,7 +130,11 @@ function createWindow() {
     minWidth: 900,
     minHeight: 600,
     icon: loadIcon(iconFile),
-    webPreferences: { nodeIntegration: false, contextIsolation: true },
+    webPreferences: {
+      nodeIntegration: false,
+      contextIsolation: true,
+      preload: path.join(__dirname, 'preload.cjs'),
+    },
     titleBarStyle: process.platform === 'darwin' ? 'hiddenInset' : 'default',
   });
 
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -82,8 +82,10 @@ ipcMain.handle('passwords:available', () => {
   try { return safeStorage.isEncryptionAvailable(); } catch { return false; }
 });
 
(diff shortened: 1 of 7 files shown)


src/components/ConnectionManager.tsx
```diff
@@ -68,9 +92,15 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
   };
 
   const applySaved = (entry: SavedConnection) => {
-    setForm({ ...entry, password: '' });
+    setForm({ ...entry, password: '', sslVerify: entry.sslVerify ?? false, savePassword: entry.savePassword ?? false });
     setAppliedSaved(entry.name);
     setSuggestOpen(false);
+    if (entry.savePassword && electronAPI) {
+      electronAPI.passwords.load(entry.name).then(pw => {
```
eugene-mobile, line 101: **Race condition: async IPC load can overwrite user typing**

The guard `p.name === entry.name` prevents writing to the wrong entry, but not to one the user has already started typing in. If the IPC round-trip is slow and the user types a password in the gap, the loaded value will silently overwrite it.

Add a check that the field is still empty:
```ts
setForm(p => p.name === entry.name && p.password === '' ? { ...p, password: pw } : p);
```
The same pattern applies to the `useEffect` that loads the initial entry's password on mount (line 63).

electron/main.cjs
```diff
@@ -52,6 +53,65 @@ function waitForServer(maxAttempts = 50) {
...
+  } catch (err) {
+    console.error('[passwords] write failed:', err);
+  }
+}
+
+ipcMain.handle('passwords:available', () => {
+  try { return safeStorage.isEncryptionAvailable(); } catch { return false; }
+});
+
+ipcMain.handle('passwords:save', (_e, name, password) => {
```
eugene-mobile, line 87: **`passwords:save` accepts an empty password**

An empty string passes the current `typeof password !== 'string'` check and gets encrypted + stored, silently overwriting a real password with an encrypted empty string. Add `|| !password` to the guard:
```js
if (typeof name !== 'string' || !name || typeof password !== 'string' || !password) return;
```

electron/main.cjs
```diff
@@ -52,6 +53,65 @@ function waitForServer(maxAttempts = 50) {
...
+  if (typeof name !== 'string' || !name || typeof password !== 'string') return;
+  if (!safeStorage.isEncryptionAvailable()) throw new Error('Encryption not available on this system');
+  const all = readPasswords();
```
eugene-mobile, line 90: **Orphaned keychain entries on connection rename**

The keychain key is the connection name. If the user renames a saved connection and reconnects with "Save password" checked, the old name's entry is never deleted — only the new name gets a new entry. These orphans accumulate silently over time.

One approach: accept an optional `oldName` argument in `passwords:save` and delete the old entry when it differs from `name`. Alternatively, document this as a known limitation for a follow-up.

src/components/ConnectionManager.tsx
```diff
@@ -41,6 +45,26 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
...
+  useEffect(() => {
+    const first = saved[0];
+    if (!first?.savePassword || !electronAPI) return;
+    let cancelled = false;
+    electronAPI.passwords.load(first.name).then(pw => {
+      if (cancelled || pw === null) return;
+      setForm(p => p.name === first.name ? { ...p, password: pw } : p);
+    }).catch(() => {});
+    return () => { cancelled = true; };
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
eugene-mobile, line 68: **eslint-disable without explanation**

The empty deps array is intentionally correct (run once on mount to pre-fill the password for the connection shown when the modal opened, not on every change to `saved`). Worth stating so a future reader doesn't "fix" it:
```ts
  // Intentional empty deps — we only want to load the password for the connection
  // that was active when the modal opened, not on every re-render.
  }, []);
```

eugene-mobile (commented):
Solid implementation — safeStorage is the right tool, the preload is minimal and correct, the browser fallback (canSavePassword = false) hides the feature cleanly, and the save-on-connect / delete-on-uncheck flow in App.tsx is well thought out. Four things worth addressing before merge.

eugene-mobile (commented):
Solid implementation — safeStorage is the right tool, the preload is minimal and correct, the browser fallback (canSavePassword = false) hides the feature cleanly, and the save-on-connect / delete-on-uncheck flow in App.tsx is well thought out. Four things worth addressing before merge.

src/components/ConnectionManager.tsx
eugene-mobile, line 101: **Race condition: async IPC load can overwrite user typing**

The guard `p.name === entry.name` prevents writing to the wrong entry, but not to one the user has already started typing in. If the IPC round-trip is slow and the user types a password in the gap, the loaded value will silently overwrite it.

Add a check that the field is still empty:
```ts
setForm(p => p.name === entry.name && p.password === '' ? { ...p, password: pw } : p);
```
The same pattern applies to the `useEffect` that loads the initial entry's password on mount (line 63).

electron/main.cjs
eugene-mobile, line 87: **`passwords:save` accepts an empty password**

An empty string passes the current `typeof password !== 'string'` check and gets encrypted + stored, silently overwriting a real password with an encrypted empty string. Add `|| !password` to the guard:
```js
if (typeof name !== 'string' || !name || typeof password !== 'string' || !password) return;
```

electron/main.cjs
eugene-mobile, line 90: **Orphaned keychain entries on connection rename**

The keychain key is the connection name. If the user renames a saved connection and reconnects with "Save password" checked, the old name's entry is never deleted — only the new name gets a new entry. These orphans accumulate silently over time.

One approach: accept an optional `oldName` argument in `passwords:save` and delete the old entry when it differs from `name`. Alternatively, document this as a known limitation for a follow-up.

src/components/ConnectionManager.tsx
eugene-mobile, line 68: **eslint-disable without explanation**

The empty deps array is intentionally correct (run once on mount to pre-fill the password for the connection shown when the modal opened, not on every change to `saved`). Worth stating so a future reader doesn't "fix" it:
```ts
  // Intentional empty deps — we only want to load the password for the connection
  // that was active when the modal opened, not on every re-render.
  }, []);
```

Issue #92, opened 2026-04-28
Refactor server/src/db.ts - use DbDriver abstraction


eugene-mobile:
﻿Part of the PostgreSQL support project (https://github.com/orgs/appspace/projects/1).

## Task
Refactor `server/src/db.ts` to use the driver abstraction instead of direct mysql2 usage.

## Acceptance Criteria
- [ ] mysql2 import removed
- [ ] Stores a `DbDriver | null` instance
- [ ] `connect()` instantiates the correct driver based on `config.type`
- [ ] Exports `getDriver()` instead of `getPool()`
- [ ] `testConnection()` delegates to the driver's `ping()`

PR #101, opened 2026-04-28
feat: PostgresDriver - pg.Pool implementation of DbDriver

### LINKED ISSUE #91: Create server/src/drivers/postgres.ts - PostgreSQL DbDriver implementation

## Task
Create `server/src/drivers/postgres.ts` using `pg.Pool`, implementing all `DbDriver` methods.

## Acceptance Criteria
- [ ] `PostgresDriver` class implements `DbDriver` interface
- [ ] `escapeIdent()` uses double-quotes
- [ ] `setSchema()` uses `SET search_path TO "schema"`
- [ ] `getSchemas()` queries `information_schema.SCHEMATA`, excluding `pg_catalog`, `information_schema`, `pg_toast*`
- [ ] `getSchema()` translates MySQL-specific SQL: `IF()` to `CASE WHEN`, `COLUMN_TYPE+EXTRA` to `udt_name+column_default`, `TABLE_ROWS` to `pg_stat_user_tables` estimate
- [ ] `getTableDdl()` reconstructs DDL from `information_schema.columns` + PKs from `table_constraints` + indexes from `pg_indexes`
- [ ] `query()` normalizes pg's `{ rows, fields }` to shared `QueryResult` shape

eugene-mobile:
## Summary
- Creates `server/src/drivers/postgres.ts` with `PostgresDriver` implementing the `DbDriver` interface (closes #91)

## What's in the driver

| Method | Implementation |
|---|---|
| `escapeIdent()` | Double-quote quoting |
| `query()` | `SET search_path` for schema switching; normalizes pg `rows`/`fields` to `QueryResult`; serializes `Buffer`→hex, `Date`→string |
| `getSchemas()` | `information_schema.schemata` excluding `pg_catalog`, `information_schema`, `pg_toast*` |
| `getSchema()` | `udt_name` for column types; `is_identity`/`nextval(` for auto-increment; `n_live_tup` for row estimates; subquery join for PK detection |
| `getTableDdl()` | Reconstructs DDL from `information_schema.columns` + `table_constraints` + `pg_indexes`; `pg_get_functiondef` for functions; `pg_get_triggerdef` for triggers |

No existing routes are changed — the driver is available but not yet wired up. That happens in the next PR (#92).

## Test plan
- [ ] `npx tsc --noEmit` passes
- [ ] `npm test` (unit suite) passes — 11/11 green

### DIFF
--- a/server/src/drivers/postgres.ts
+++ b/server/src/drivers/postgres.ts
@@ -1,0 +1,321 @@
+import pg from 'pg';
+import type { DbDriver, ConnectionConfig, QueryResult, ColumnMeta, ColumnInfo, SchemaInfo } from './interface.js';
+
+export class PostgresDriver implements DbDriver {
+  private pool: pg.Pool;
+
+  constructor(config: ConnectionConfig) {
+    this.pool = new pg.Pool({
+      host: config.host,
+      port: config.port,
+      user: config.user,
+      password: config.password,
+      database: config.database || undefined,
+      ssl: config.ssl === 'verify-full' ? { rejectUnauthorized: true }
+         : config.ssl === 'require'     ? { rejectUnauthorized: false }
+         : undefined,
+      max: 5,
+      connectionTimeoutMillis: 10_000,
+    });
+  }
+
+  escapeIdent(s: string): string {
+    return '"' + s.replace(/"/g, '') + '"';
+  }
+
+  async ping(): Promise<void> {
+    const client = await this.pool.connect();
+    try {
+      await client.query('SELECT 1');
+    } finally {
+      client.release();
+    }
+  }
+
+  async end(): Promise<void> {
+    await this.pool.end();
+  }
+
+  async query(sql: string, params?: unknown[], schema?: string): Promise<QueryResult> {
+    const client = await this.pool.connect();
+    try {
+      if (schema) {
+        await client.query(`SET search_path TO ${this.escapeIdent(schema)}`);
+      }
+
+      const result = await client.query({ text: sql, values: params?.length ? params : undefined });
+
+      // Non-SELECT (INSERT, UPDATE, DELETE, DDL, etc.)
+      if (!result.fields || result.fields.length === 0) {
+        return {
+          rows: [],
+          columnMeta: [],
+          affectedRows: result.rowCount ?? 0,
+          insertId: null,
+        };
+      }
+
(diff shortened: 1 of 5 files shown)


eugene-mobile (commented):
Two correctness bugs: `escapeIdent` corrupts identifiers and `search_path` leaks between pool connections. Everything else looks solid.

server/src/drivers/postgres.ts
```diff
@@ -0,0 +1,321 @@
...
+      ssl: config.ssl === 'verify-full' ? { rejectUnauthorized: true }
+         : config.ssl === 'require'     ? { rejectUnauthorized: false }
+         : undefined,
+      max: 5,
+      connectionTimeoutMillis: 10_000,
+    });
+  }
+
+  escapeIdent(s: string): string {
+    return '"' + s.replace(/"/g, '') + '"';
```
eugene-mobile, line 23: **`escapeIdent` strips quotes instead of escaping them**

The current implementation removes `"` characters:
```ts
return '"' + s.replace(/"/g, '') + '"';
```
PostgreSQL's SQL standard requires that a literal double-quote *inside* a double-quoted identifier be represented as `""`. Stripping them instead silently corrupts any identifier that contains a quote (e.g. `foo"bar` becomes `"foobar"` — a different identifier). This also means the escaping is technically wrong for the `SET search_path TO ...` call below.

Fix:
```ts
return '"' + s.replace(/"/g, '""') + '"';
```

server/src/drivers/postgres.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  async end(): Promise<void> {
+    await this.pool.end();
+  }
+
+  async query(sql: string, params?: unknown[], schema?: string): Promise<QueryResult> {
+    const client = await this.pool.connect();
+    try {
+      if (schema) {
+        await client.query(`SET search_path TO ${this.escapeIdent(schema)}`);
```
eugene-mobile, line 43: **`search_path` leaks to future pool checkouts**

`SET search_path TO ...` modifies session-level state on this connection. When the client is released back to the pool (`client.release()` on line 89), the `search_path` is still set. The next caller that checks out the same connection will silently run queries against the wrong schema.

Fix — reset in the `finally` block before releasing:
```ts
} finally {
  if (schema) await client.query('RESET search_path');
  client.release();
}
```

Issue #97, opened 2026-04-28
Update insertRow, updateCell, deleteRow routes - use driver.escapeIdent()


eugene-mobile:
﻿Part of the PostgreSQL support project (https://github.com/orgs/appspace/projects/1).

## Task
Update `server/src/routes/insertRow.ts`, `updateCell.ts`, and `deleteRow.ts` to use dialect-agnostic identifier escaping.

## Acceptance Criteria
- [ ] Hardcoded backtick escaping replaced with `driver.escapeIdent()` in all three files
- [ ] Works correctly for both MySQL (backticks) and PostgreSQL (double-quotes)

eugene-mobile:
Completed by #102 (merged 2026-04-29). All acceptance criteria met on main:
- All routes now delegate to the `DbDriver` abstraction (`getDriver().query() / getSchemas() / getSchema() / getTableDdl()`).
- Identifier escaping uses `driver.escapeIdent()` (backticks for MySQL, double-quotes for Postgres).
- `connect.ts` validates `type` strictly, defaults port per type, and `friendlyConnectError` includes Postgres SQLSTATE codes (`28P01`, `3D000`, `42501`, `28000`, `08006`).

Issue #96, opened 2026-04-28
Update server/src/routes/tableDdl.ts - delegate to DbDriver


eugene-mobile:
﻿Part of the PostgreSQL support project (https://github.com/orgs/appspace/projects/1).

## Task
Update `server/src/routes/tableDdl.ts` to use the driver abstraction.

## Acceptance Criteria
- [ ] `SHOW CREATE TABLE` replaced with `driver.getTableDdl()`
- [ ] Works for both MySQL and PostgreSQL (Postgres reconstructs DDL from information_schema + pg_indexes)

eugene-mobile:
Completed by #102 (merged 2026-04-29). All acceptance criteria met on main:
- All routes now delegate to the `DbDriver` abstraction (`getDriver().query() / getSchemas() / getSchema() / getTableDdl()`).
- Identifier escaping uses `driver.escapeIdent()` (backticks for MySQL, double-quotes for Postgres).
- `connect.ts` validates `type` strictly, defaults port per type, and `friendlyConnectError` includes Postgres SQLSTATE codes (`28P01`, `3D000`, `42501`, `28000`, `08006`).

Issue #94, opened 2026-04-28
Update server/src/routes/query.ts - use DbDriver abstraction


eugene-mobile:
﻿Part of the PostgreSQL support project (https://github.com/orgs/appspace/projects/1).

## Task
Update `server/src/routes/query.ts` to use the driver abstraction.

## Acceptance Criteria
- [ ] Replace `getPool()` and mysql2 type imports with `getDriver().query()`
- [ ] Replace `USE schema` with `driver.setSchema(schema)`
- [ ] Column metadata sourced from normalized `QueryResult` ? mysql2-specific flag bitmask logic removed

eugene-mobile:
Completed by #102 (merged 2026-04-29). All acceptance criteria met on main:
- All routes now delegate to the `DbDriver` abstraction (`getDriver().query() / getSchemas() / getSchema() / getTableDdl()`).
- Identifier escaping uses `driver.escapeIdent()` (backticks for MySQL, double-quotes for Postgres).
- `connect.ts` validates `type` strictly, defaults port per type, and `friendlyConnectError` includes Postgres SQLSTATE codes (`28P01`, `3D000`, `42501`, `28000`, `08006`).

Issue #93, opened 2026-04-28
Update server/src/routes/connect.ts - support MySQL and PostgreSQL


eugene-mobile:
﻿Part of the PostgreSQL support project (https://github.com/orgs/appspace/projects/1).

## Task
Update `server/src/routes/connect.ts` to support both MySQL and PostgreSQL connections.

## Acceptance Criteria
- [ ] Accepts `type: 'mysql' | 'postgres'` in request body
- [ ] Default port is 3306 for MySQL, 5432 for PostgreSQL
- [ ] Postgres error codes mapped: `28P01` = wrong password, `3D000` = database not found, `42501` = permission denied
- [ ] MySQL-specific strings removed from generic error messages

eugene-mobile:
Completed by #102 (merged 2026-04-29). All acceptance criteria met on main:
- All routes now delegate to the `DbDriver` abstraction (`getDriver().query() / getSchemas() / getSchema() / getTableDdl()`).
- Identifier escaping uses `driver.escapeIdent()` (backticks for MySQL, double-quotes for Postgres).
- `connect.ts` validates `type` strictly, defaults port per type, and `friendlyConnectError` includes Postgres SQLSTATE codes (`28P01`, `3D000`, `42501`, `28000`, `08006`).

PR #102, opened 2026-04-29
feat: wire all routes to DbDriver abstraction

### LINKED ISSUE #92: Refactor server/src/db.ts - use DbDriver abstraction

eugene-mobile:
## Summary

- Replaced all direct mysql2 pool usage in routes with the DbDriver abstraction (getDriver())
- Added rowLimitClause(n) to DbDriver interface so MySQL emits LIMIT 1 on DML while Postgres omits it
- Rewrote mcp.ts to use driver methods (getSchemas, getSchema, query) - fully dialect-agnostic
- Updated dropTable.ts, query.ts, schema.ts, tableDdl.ts, insertRow.ts, updateCell.ts, deleteRow.ts
- Added Postgres integration test suite (19 tests) with separate vitest config and Docker Compose service
- Excluded integration tests from the default npm test run

Closes #92, #93, #94, #95, #96, #97

## Test plan

- [ ] npm test in server/ passes (7/7 unit tests)
- [ ] Postgres integration tests pass (19/19)
- [ ] MySQL integration tests pass
- [ ] TypeScript compiles cleanly

### DIFF
--- a/server/src/drivers/postgres.ts
+++ b/server/src/drivers/postgres.ts
@@ -1,5 +1,5 @@
 import pg from 'pg';
-import type { DbDriver, ConnectionConfig, QueryResult, ColumnMeta, ColumnInfo, SchemaInfo } from './interface.js';
+import type { DbDriver, ConnectionConfig, QueryResult, ColumnMeta, ColumnInfo, SchemaInfo, TableInfo } from './interface.js';
 
 export class PostgresDriver implements DbDriver {
   private pool: pg.Pool;
@@ -42,12 +42,15 @@ export class PostgresDriver implements DbDriver {
 
   async query(sql: string, params?: unknown[], schema?: string): Promise<QueryResult> {
     const client = await this.pool.connect();
+    let searchPathSet = false;
     try {
       if (schema) {
         await client.query(`SET search_path TO ${this.escapeIdent(schema)}`);
+        searchPathSet = true;
       }
 
-      // Convert MySQL-style ? placeholders to Postgres $N so DML routes stay uniform
+      // Naive ?→$N rewrite: assumes machine-generated SQL when params is non-empty.
+      // User-authored SQL must be passed without params (the rewrite is skipped then).
       let pgSql = sql;
       if (params?.length) {
         let n = 0;
@@ -96,6 +99,11 @@ export class PostgresDriver implements DbDriver {
 
       return { rows: serializedRows, columnMeta };
     } finally {
+      // pg.Pool reuses clients without resetting session state, so search_path
+      // would leak to the next caller on this same client.
+      if (searchPathSet) {
+        try { await client.query('SET search_path TO DEFAULT'); } catch { /* fall through to release */ }
+      }
       client.release();
     }
   }
@@ -196,6 +204,70 @@ export class PostgresDriver implements DbDriver {
         views: viewsRes.rows.map(v => v.name),
         procedures: procsRes.rows.map(p => p.name),
         triggers: triggersRes.rows.map(t => t.name),
+      };
+    } finally {
+      client.release();
+    }
+  }
+
+  async getTable(schema: string, table: string): Promise<TableInfo | null> {
+    const client = await this.pool.connect();
+    try {
+      const tableRes = await client.query<{ name: string; row_count: string }>(
+        `SELECT t.table_name AS name,
+                COALESCE(s.n_live_tup, 0)::text AS row_count
+         FROM information_schema.tables t
+         LEFT JOIN pg_stat_user_tables s
+           ON s.schemaname = t.table_schema AND s.relname = t.table_name
+         WHERE t.table_schema = $1 AND t.table_name = $2 AND t.table_type = 'BASE TABLE'`,
+        [schema, table],
+      );
+      if (tableRes.rows.length === 0) return null;
+
+      const colsRes = await client.query<{
+        col: string; col_type: string; data_type: string;
+        is_pk: string; nullable: string; col_default: string | null; extra: string;
+      }>(
+        `SELECT c.column_name AS col,
+                c.udt_name AS col_type,
+                c.data_type AS data_type,
+                CASE WHEN pk.column_name IS NOT NULL THEN '1' ELSE '0' END AS is_pk,
+                CASE WHEN c.is_nullable = 'YES' THEN '1' ELSE '0' END AS nullable,
+                c.column_default AS col_default,
+                CASE WHEN c.column_default LIKE 'nextval(%' OR c.is_identity = 'YES'
+                     THEN 'auto_increment' ELSE '' END AS extra
+         FROM information_schema.columns c
+         LEFT JOIN (
+           SELECT kcu.table_name, kcu.column_name, kcu.table_schema
+           FROM information_schema.key_column_usage kcu
+           JOIN information_schema.table_constraints tc
+             ON tc.constraint_name = kcu.constraint_name
+             AND tc.table_schema = kcu.table_schema
+             AND tc.table_name = kcu.table_name
+             AND tc.constraint_type = 'PRIMARY KEY'
+           WHERE kcu.table_schema = $1 AND kcu.table_name = $2
+         ) pk ON pk.column_name = c.column_name
+         WHERE c.table_schema = $1 AND c.table_name = $2
+         ORDER BY c.ordinal_position`,
+        [schema, table],
+      );
+
+      const cols: ColumnInfo[] = colsRes.rows.map(r => ({
+        name: r.col,
+        type: r.col_type,
+        dataType: (r.data_type ?? '').toLowerCase(),
+        pk: r.is_pk === '1',
+        nullable: r.nullable === '1',
+        default: r.col_default ?? null,
+        autoIncrement: r.extra.includes('auto_increment'),
+        comment: '',
+      }));
+
+      return {
+        name: tableRes.rows[0].name,
+        rows: Number(tableRes.rows[0].row_count),
+        comment: '',
+        columns: cols,
       };
     } finally {
       client.release();
--- a/server/src/drivers/postgres.ts
+++ b/server/src/drivers/postgres.ts
@@ -23,6 +23,10 @@ export class PostgresDriver implements DbDriver {
     return '"' + s.replace(/"/g, '') + '"';
   }
 
+  rowLimitClause(_n: number): string {
+    return '';
+  }
+
   async ping(): Promise<void> {
     const client = await this.pool.connect();
     try {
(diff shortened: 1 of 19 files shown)


server/src/drivers/postgres.ts
```diff
@@ -43,7 +47,13 @@ export class PostgresDriver implements DbDriver {
         await client.query(`SET search_path TO ${this.escapeIdent(schema)}`);
```
eugene-mobile, line 48: **Connection state leak (correctness).** `pg.Pool` does not reset session state when a client is `release()`'d — `SET search_path` persists on that pooled connection and will be inherited by the next caller. So a sequence of `query(..., schema='a')` then `query(..., /* no schema */)` can run against schema `a` instead of the configured default. Two ways to fix:

1. Always restore the default after the user query: `SET search_path TO DEFAULT` in a `try/finally`.
2. Or use `client.release(true)` to destroy the connection on release (heavier).

The MySQL driver has the analogous problem with `USE`, but every route that takes a schema also passes it on every call, so leaked state happens to be overwritten. Postgres doesn't get that lucky once a route omits the schema.

server/src/drivers/postgres.ts
```diff
@@ -43,7 +47,13 @@ export class PostgresDriver implements DbDriver {
...
       }
 
-      const result = await client.query({ text: sql, values: params?.length ? params : undefined });
+      // Convert MySQL-style ? placeholders to Postgres $N so DML routes stay uniform
+      let pgSql = sql;
+      if (params?.length) {
+        let n = 0;
+        pgSql = sql.replace(/\?/g, () => `$${++n}`);
```
eugene-mobile, line 57: **Naive `?` substitution corrupts string literals.** `sql.replace(/\?/g, ...)` rewrites every `?` in the SQL, including ones inside string literals, identifiers, dollar-quoted bodies, and comments. For example, `INSERT INTO t (q) VALUES ('what?')` with one param becomes `... VALUES ('what$1')`, which silently injects the placeholder into the literal.

For the route-generated DML (`insertRow`, `updateCell`, `deleteRow`) this happens to be safe because those SQL strings never contain user content as literal `?`. But `postQuery` (and the MCP `execute_*` tools) pass user-authored SQL through unchanged, so a Postgres user typing `SELECT 'foo?bar'` will hit this. At minimum, document that the driver-level rewrite assumes machine-generated SQL, and avoid passing user SQL through it (or skip the rewrite when `params` is empty — already done — and never pass params for user SQL paths).

server/src/routes/updateCell.ts
```diff
@@ -35,18 +32,18 @@ export const postUpdateCell: RequestHandler = async (req, res) => {
...
+  const whereClause = where.map(w => `${driver.escapeIdent(w.column)} = ?`).join(' AND ');
+  const sql = `UPDATE ${qualifiedTable} SET ${driver.escapeIdent(column)} = ? WHERE ${whereClause}${driver.rowLimitClause(1)}`;
   const params = [value === undefined ? null : value, ...where.map(w => w.value)];
 
   try {
-    const pool = getPool();
-    const [result] = await pool.query(sql, params) as [ResultSetHeader, unknown];
-    res.json({ affectedRows: result.affectedRows ?? 0, changedRows: (result as unknown as { changedRows?: number }).changedRows ?? 0, sql });
+    const result = await driver.query(sql, params);
+    res.json({ affectedRows: result.affectedRows ?? 0, sql });
```
eugene-mobile, line 46: **API contract regression.** The previous response included `changedRows` (distinct from `affectedRows` in MySQL — `affectedRows` counts matched rows, `changedRows` counts rows whose values actually differed). This commit drops it because `QueryResult` no longer surfaces it.

`src/api.ts:111` still declares the response type as `{ affectedRows: number; changedRows: number; sql: string }`, so the frontend will silently get `undefined` for `changedRows`. Either restore `changedRows` on `QueryResult` (MySQL-only, undefined for Postgres) and keep returning it here, or update `api.ts` to drop the field. The same applies to `mcp.ts` `execute_write` which previously returned `changedRows` too.

server/src/mcp.ts
```diff
@@ -131,36 +96,12 @@ export function buildMcpServer(): McpServer {
...
-                  EXTRA AS extra,
-                  COLUMN_COMMENT AS comment
-           FROM information_schema.COLUMNS
-           WHERE TABLE_SCHEMA = ? AND TABLE_NAME = ?
-           ORDER BY ORDINAL_POSITION`,
-          [schema, table],
-        );
-        if (columns.length === 0) {
-          return toolError(`Table \`${schema}\`.\`${table}\` not found.`);
+        const info = await getDriver().getSchema(schema);
```
eugene-mobile: **Performance regression.** `describe_table` previously issued one targeted query against `information_schema.COLUMNS` filtered by `(schema, table)`. It now calls `getDriver().getSchema(schema)` which loads tables, *all* columns for *all* tables, views, procedures, and triggers for the entire schema, then filters in JS. On schemas with hundreds of tables this is a large unnecessary fan-out. Consider adding a `getTable(schema, table)` method on `DbDriver` that does a targeted lookup.

server/src/routes/tableDdl.ts
```diff
@@ -23,36 +20,10 @@ export const getTableDdl: RequestHandler = async (req, res) => {
...
-    const ddl = (
-      row['Create Table'] ??
-      row['Create View'] ??
-      row['Create Procedure'] ??
-      row['SQL Original Statement'] ??   // SHOW CREATE TRIGGER
-      ''
-    ) as string;
+    const ddl = await getDriver().getTableDdl(schema, table, type as 'table' | 'view' | 'procedure' | 'trigger');
     res.json({ ddl });
   } catch (err) {
```
eugene-mobile, line 25: **Fragile error-class detection.** Mapping 404 vs 400 by string-matching `'No DDL returned'` in the thrown message couples this route to the exact wording of the driver's error. If a driver implementer rewords the message, this silently degrades to 400. Consider throwing a typed error (e.g. `class NotFoundError extends Error`) from drivers and `instanceof`-checking here instead.

server/src/routes/connect.ts
```diff
@@ -37,32 +53,36 @@ function friendlyConnectError(err: unknown, host: string, port: number): string
...
     ssl?: 'require' | 'verify-full';
+    type?: DbType;
   };
 
   if (!host || !user) {
     res.status(400).json({ error: 'host and user are required.' });
     return;
   }
 
+  const dbType: DbType = type === 'postgres' ? 'postgres' : 'mysql';
```
eugene-mobile: **Silent fallback hides bad input.** `type === 'postgres' ? 'postgres' : 'mysql'` accepts anything that isn't literally `'postgres'` (including `'mariadb'`, `'sqlite'`, `42`, etc.) and silently treats it as MySQL. Prefer an explicit allowlist that 400s on unknown values:

```ts
if (type !== undefined && type !== 'mysql' && type !== 'postgres') {
  res.status(400).json({ error: `Unsupported db type: ${type}` });
  return;
}
const dbType: DbType = type ?? 'mysql';
```

This logic is also duplicated verbatim in `postTestConnect` — worth extracting a small `parseConnectBody(req)` helper.

server/src/routes/query.test.ts
```diff
@@ -35,93 +20,80 @@ function makeApp() {
   return app;
 }
```
eugene-mobile, line 22: **Test gap.** The previous test suite asserted that `release()` was called on every path (success, query throws, USE throws). With `withSchema` removed, those assertions are gone — but the responsibility for connection release didn't disappear, it moved into `MysqlDriver.query` and `PostgresDriver.query`. There are no unit tests for those drivers covering the error-path release. Recommend adding `mysql.test.ts` / `postgres.test.ts` (mocking the underlying pool) that assert `client.release()` runs when the inner `query` rejects.

server/src/routes/query.ts
```diff
@@ -11,72 +10,26 @@ export const postQuery: RequestHandler = async (req, res) => {
...
-            // Assumes the MySQL server is also UTC — displayed time will be wrong otherwise.
-            out[col] = val.toISOString().replace('T', ' ').replace(/\.\d{3}Z$/, '');
-          } else if (typeof val === 'bigint') {
-            out[col] = val.toString();
-          } else {
-            out[col] = val;
-          }
-        }
-        return out;
+    if (result.columnMeta.length === 0) {
```
eugene-mobile, line 17: **Behavior change worth confirming.** The route now decides "this was DML" purely from `result.columnMeta.length === 0`. A `SELECT` that returns zero columns is essentially impossible, so this is fine for MySQL. For Postgres, however, statements like `CREATE TABLE` also land here with empty `columnMeta` and `affectedRows: 0` — the response will set `affectedRows: 0, insertId: null`, which is correct but indistinguishable from a no-op DML. Consider whether the frontend cares to distinguish DDL-success from "DML matched 0 rows".

eugene-mobile (commented):
Solid abstraction work overall. A few correctness issues worth fixing before merge — most notably the Postgres search_path leak across pooled clients, the naive ?->$N substitution that mangles literal `?` characters in user SQL, and the silently-dropped `changedRows` field that the frontend API contract still declares.

eugene-mobile (commented):
Re-review of fix commit bf1cd7a — all 5 in-scope review comments are adequately resolved, and the 3 explicitly-deferred items (?→$N rewrite, 'No DDL returned' string match, DDL vs zero-row-DML) are acceptable deferrals.

**Verified fixes:**

1. **Postgres search_path leak** (`drivers/postgres.ts:43-108`) — `searchPathSet` flag is set only after `SET` succeeds; `finally` resets to `DEFAULT` only when needed, swallows reset errors so `release()` always runs. Bullet-proof against SET failing, user query rejecting, and the reset call itself throwing.

2. **`changedRows` dropped consistently** — `src/api.ts:111` no longer declares `changedRows`. No other callers in `src/` or tests reference it. `updateCell.ts` route response matches.

3. **db `type` validation** (`routes/connect.ts:73-87`) — `parseConnectBody` does an explicit allowlist (`type !== 'mysql' && type !== 'postgres'`), returns 400 with a clear message. Used by both `postConnect` and `postTestConnect` — duplication removed.

4. **`getTable()` driver method** — Added to `DbDriver` interface; `MysqlDriver.getTable` filters on `TABLE_SCHEMA = ? AND TABLE_NAME = ?`, `PostgresDriver.getTable` filters on `table_schema = $1 AND table_name = $2`; both return `null` when missing and `release()` in `finally` on every path. `mcp.ts describe_table` now calls `getDriver().getTable(...)` directly — the fan-out concern is gone.

5. **Driver release-on-error tests** — `mysql.test.ts` covers success / query-rejects / USE-rejects. `postgres.test.ts` covers success / query-rejects / SET-rejects / `SET search_path TO DEFAULT` issued before release on both happy and user-query-rejects paths. All 15 tests pass locally.

**Recommendation: approve.** No new issues introduced by the follow-up commit.

Issue #99, opened 2026-04-28
Update ConnectionManager.tsx - add DB type selector and dynamic port


eugene-mobile:
﻿Part of the PostgreSQL support project (https://github.com/orgs/appspace/projects/1).

## Task
Update `src/components/ConnectionManager.tsx` to support selecting the database type.

## Acceptance Criteria
- [ ] DB type toggle (MySQL / PostgreSQL) added at the top of the connection form
- [ ] Port field auto-updates when type changes, if it still holds the previous default (3306 <-> 5432)
- [ ] `type` included in connect and test-connect API payloads

PR #103, opened 2026-04-29
feat: add DB type selector to connection UI

### LINKED ISSUE #98: Update src/savedConnections.ts - add type field to SavedConnection

## Task
Update `src/savedConnections.ts` to include a `type` field on saved connections.

## Acceptance Criteria
- [ ] `type: 'mysql' | 'postgres'` added to the `SavedConnection` type
- [ ] Existing saved connections without `type` default to `'mysql'` for backwards compatibility

eugene-mobile:
## Summary

- Added `type: "mysql" | "postgres"` field to `SavedConnection` (optional, defaults to `"mysql"` for backward compatibility with existing saved connections)
- Added `type` to `ConnectionForm` in `ConnectionManager`
- Connection modal now shows **MySQL / PostgreSQL** toggle buttons at the top of the form
- Switching DB type auto-updates the port when it is still the default (3306 to 5432 and back)
- Modal subtitle updates to reflect the selected DB type
- `api.ts` forwards `type` in connect and testConnection request bodies
- `server/src/routes/connect.ts` extracts `type` from the request and passes it to `connect()` and `testConnection()`
- `server/src/db.ts` `ConnectionConfig` accepts optional `type` field
- `server/vitest.config.ts` excludes integration tests from the default `npm test` run

Closes #98, #99

## Test plan

- [ ] TypeScript compiles cleanly (frontend + server)
- [ ] `npm test` in `server/` passes (11/11 unit tests)
- [ ] MySQL/PostgreSQL toggle appears in connection modal
- [ ] Switching to PostgreSQL changes port from 3306 to 5432 (and back)
- [ ] Subtitle changes to "Connect to a PostgreSQL server" when Postgres is selected
- [ ] Connecting to MySQL still works end-to-end

### DIFF
--- a/src/components/ConnectionManager.tsx
+++ b/src/components/ConnectionManager.tsx
@@ -95,14 +95,17 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
     setTestResult(null);
   };
 
-  const setDbType = (t: 'mysql' | 'postgres') => {
-    setForm(p => ({
-      ...p,
-      type: t,
-      port: p.port === '3306' || p.port === '5432' || p.port === ''
-        ? (t === 'mysql' ? '3306' : '5432')
-        : p.port,
-    }));
+  const defaultPort = (type: 'mysql' | 'postgres') => type === 'postgres' ? '5432' : '3306';
+  const setDbType = (next: 'mysql' | 'postgres') => {
+    setForm(p => {
+      const prevDefault = defaultPort(p.type);
+      const portIsPrevDefault = p.port === '' || p.port === prevDefault;
+      return {
+        ...p,
+        type: next,
+        port: portIsPrevDefault ? defaultPort(next) : p.port,
+      };
+    });
     setTestResult(null);
   };
 
--- a/src/components/ConnectionManager.tsx
+++ b/src/components/ConnectionManager.tsx
@@ -7,6 +7,7 @@ import { electronAPI } from '../electronAPI';
 
 export interface ConnectionForm {
   name: string;
+  type: 'mysql' | 'postgres';
   host: string;
   port: string;
   user: string;
@@ -30,9 +31,10 @@ interface ConnectionManagerProps {
 export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t }: ConnectionManagerProps) {
   const [saved, setSaved] = useState<SavedConnection[]>(() => listSavedConnections());
   const initialForm: ConnectionForm = saved[0]
-    ? { ...saved[0], password: '', sslVerify: saved[0].sslVerify ?? false, savePassword: saved[0].savePassword ?? false }
+    ? { ...saved[0], type: saved[0].type ?? 'mysql', password: '', sslVerify: saved[0].sslVerify ?? false, savePassword: saved[0].savePassword ?? false }
     : {
         name: 'Local MySQL',
+        type: 'mysql',
         host: import.meta.env['VITE_DEFAULT_HOST'] ?? 'localhost',
         port: import.meta.env['VITE_DEFAULT_PORT'] ?? '3306',
         user: import.meta.env['VITE_DEFAULT_USER'] ?? 'root',
@@ -93,8 +95,19 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
     setTestResult(null);
   };
 
+  const setDbType = (t: 'mysql' | 'postgres') => {
+    setForm(p => ({
+      ...p,
+      type: t,
+      port: p.port === '3306' || p.port === '5432' || p.port === ''
+        ? (t === 'mysql' ? '3306' : '5432')
+        : p.port,
+    }));
+    setTestResult(null);
+  };
+
   const applySaved = (entry: SavedConnection) => {
-    setForm({ ...entry, password: '', sslVerify: entry.sslVerify ?? false, savePassword: entry.savePassword ?? false });
+    setForm({ ...entry, type: entry.type ?? 'mysql', password: '', sslVerify: entry.sslVerify ?? false, savePassword: entry.savePassword ?? false });
     setAppliedSaved(entry.name);
     setSuggestOpen(false);
     if (entry.savePassword && electronAPI) {
@@ -160,7 +173,7 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
           </svg>
           <div>
             <div style={s.title}>New connection</div>
-            <div style={s.subtitle}>Connect to a MySQL server</div>
+            <div style={s.subtitle}>Connect to a {form.type === 'postgres' ? 'PostgreSQL' : 'MySQL'} server</div>
           </div>
         </div>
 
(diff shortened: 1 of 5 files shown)


eugene-mobile (commented):
Request changes. The headline feature does not actually work: although the UI now lets users pick PostgreSQL and the route forwards `type` into `connect()`, `db.ts` ignores it and always creates a `mysql.createPool` / `mysql.createConnection`. So selecting Postgres in the modal will hit a Postgres server with the MySQL wire protocol. Also worth resolving the merge conflict against main and adding a unit test for the new `type` plumbing.

Separately, several smaller issues noted inline. The `vitest.config.ts` exclude is a good fix — `*.test.ts` was previously matching the integration files too.

server/src/db.ts
```diff
@@ -3,6 +3,7 @@ import type { PoolConnection } from 'mysql2/promise';
 import { resetMcpState } from './mcp-state.js';
 
 interface ConnectionConfig {
+  type?: 'mysql' | 'postgres';
```
eugene-mobile: Blocking: this is the only place where `type` actually matters and it's unused. `connect()` below unconditionally calls `mysql.createPool(...)` and `testConnection()` calls `mysql.createConnection(...)`, so picking PostgreSQL in the UI ends up speaking the MySQL protocol to a Postgres port. Either (a) wire this into the existing `DbDriver` abstraction in `drivers/interface.ts` (which is what #102 / `pr-102` is doing — coordinate with that PR to avoid duplicate work and a worse conflict), or (b) leave the field out of the UI until the backend can honor it. Right now the feature is a façade.

Note also: there are now two separate `ConnectionConfig` interfaces in the tree (this one, and `drivers/interface.ts` which makes `type` *required*) — they will need to be reconciled.

server/src/routes/connect.ts
```diff
@@ -57,12 +58,14 @@ export const postConnect: RequestHandler = async (req, res) => {
     return;
   }
 
+  const dbType = type === 'postgres' ? 'postgres' : 'mysql';
```
eugene-mobile: Soft: `type === 'postgres' ? 'postgres' : 'mysql'` silently coerces any other value (including a typo or future `'sqlite'`) to MySQL, which can mask client bugs. Prefer an explicit allowlist + 400 on unknown values, e.g.:

```ts
if (type !== undefined && type !== 'mysql' && type !== 'postgres') {
  res.status(400).json({ error: `Unsupported db type: ${type}` });
  return;
}
const dbType = type ?? 'mysql';
```

Same comment applies to the duplicated block in `postTestConnect` at line 93 — worth factoring out a small `parseConnectBody(req.body)` helper now that the shape is non-trivial.

server/src/routes/connect.ts
```diff
@@ -57,12 +58,14 @@ export const postConnect: RequestHandler = async (req, res) => {
...
+  const defaultPort = dbType === 'postgres' ? 5432 : 3306;
   try {
-    const effectivePort = Number(port) || 3306;
-    await connect({ host, port: effectivePort, user, password, database, ssl });
+    const effectivePort = Number(port) || defaultPort;
+    await connect({ type: dbType, host, port: effectivePort, user, password, database, ssl });
     res.json({ ok: true, connectionName: `${user}@${host}:${effectivePort}` });
   } catch (err) {
-    res.status(400).json({ error: friendlyConnectError(err, host, Number(port) || 3306) });
+    res.status(400).json({ error: friendlyConnectError(err, host, Number(port) || defaultPort) });
```
eugene-mobile: Once `type: 'postgres'` is honored, every error message produced by `friendlyConnectError` will still say "MySQL" — `ECONNREFUSED` will tell a Postgres user to check their MySQL server. The error codes themselves also diverge (Postgres surfaces `28P01`, `3D000`, etc. via `pg`, not `ER_ACCESS_DENIED_ERROR`). Either parameterize the messages on `dbType` or move the lookup behind the driver. Not blocking for *this* PR if you accept that Postgres support isn't actually shipped here, but please file a follow-up.

src/components/ConnectionManager.tsx
```diff
@@ -93,8 +95,19 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
     setTestResult(null);
   };
 
+  const setDbType = (t: 'mysql' | 'postgres') => {
```
eugene-mobile: Nit: parameter `t` shadows the outer `t: Theme` prop. The handler doesn't reference the theme so it works today, but it's a footgun if anyone later adds e.g. `background: t.accent` inside this closure. Rename to `next` / `nextType` (`setDbType = (next: 'mysql' | 'postgres') => …`).

src/components/ConnectionManager.tsx
```diff
@@ -93,8 +95,19 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
...
+    setForm(p => ({
+      ...p,
+      type: t,
+      port: p.port === '3306' || p.port === '5432' || p.port === ''
+        ? (t === 'mysql' ? '3306' : '5432')
```
eugene-mobile: Soft: this overwrites the user's manually-typed port whenever it happens to equal `3306` or `5432`. E.g. user picks MySQL, types `5432` because they're running MySQL on a non-standard port, then toggles to Postgres and back — their `5432` gets clobbered to `3306`. Safer rule: only auto-fill if the current port equals the *previous* type's default (so track the previous type), or if the field is empty. Minor, but easy to get right.

server/src/routes/connect.ts
```diff
@@ -57,12 +58,14 @@ export const postConnect: RequestHandler = async (req, res) => {
     return;
   }
 
+  const dbType = type === 'postgres' ? 'postgres' : 'mysql';
+  const defaultPort = dbType === 'postgres' ? 5432 : 3306;
   try {
-    const effectivePort = Number(port) || 3306;
-    await connect({ host, port: effectivePort, user, password, database, ssl });
+    const effectivePort = Number(port) || defaultPort;
+    await connect({ type: dbType, host, port: effectivePort, user, password, database, ssl });
```
eugene-mobile: Test gap: the PR description claims 11/11 tests pass but adds no coverage for the new behavior. At minimum, a unit test asserting `connect()` is called with `type: 'postgres'` when the body contains `type: 'postgres'`, and `type: 'mysql'` (or the default) otherwise. Without it, regressions in the type plumbing will go undetected — and given that the backend currently ignores the field anyway, this is the only signal that would catch the gap flagged in `db.ts`.

src/components/ConnectionManager.tsx
```diff
@@ -93,8 +95,22 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
...
   };
 
+  const defaultPort = (type: 'mysql' | 'postgres') => type === 'postgres' ? '5432' : '3306';
+  const setDbType = (next: 'mysql' | 'postgres') => {
+    setForm(p => {
+      const prevDefault = defaultPort(p.type);
+      const portIsPrevDefault = p.port === '' || p.port === prevDefault;
+      return {
+        ...p,
+        type: next,
```
eugene-mobile, line 105: Non-blocking — the heuristic still loses a deliberately-typed port across a round-trip. Repro: start on MySQL with port `3306`, the user types `5432` (e.g. their MySQL happens to listen there), switch to Postgres (port preserved as `5432` — good), then switch back to MySQL. Now `prevDefault === '5432'` and `p.port === '5432'`, so `portIsPrevDefault` is true and the typed value is replaced with `3306`. Inherent to value-based detection without a separate `userTouchedPort` flag. Fine to leave as-is given how unlikely the scenario is, but worth a TODO if you want to fully honor user-typed values.

eugene-mobile (commented):
Recommendation: APPROVE (posted as COMMENT because GitHub forbids approving one's own PR via the API).

The original blockers from my last review are resolved:
- Backend type-to-driver wiring, strict `type` allowlist (`parseConnectBody`), and Postgres-aware `friendlyConnectError` (SQLSTATE codes `28P01`, `3D000`, `42501`, `28000`, `08006`) all landed via #102 and the rebase preserved them. `db.ts` `makeDriver()` correctly switches on `config.type`.
- `setDbType`'s parameter is renamed `next`; no longer shadows the `t: Theme` prop. Closure references (`p.type`, `defaultPort(p.type)`) are correct.
- Original PR scope is intact post-rebase: toggle UI, `SavedConnection.type`, `App.tsx` save path forwarding `type`, and `api.ts` typing `type` on both `connect` and `testConnection`.
- `server/src/routes/connect.test.ts` genuinely covers the plumbing: it asserts the exact `type`/`port` fields passed to `connect()` and `testConnection()`, exercises both default and explicit ports, and verifies unknown types short-circuit to 400 before either downstream call. Mocks of `db.js` are scoped sensibly.

One non-blocking nit on the port-preserve heuristic — see inline.

PR #104, opened 2026-04-29
Dismiss Update cell / Delete row confirmations on Escape

### LINKED ISSUE #38: Escape key does not dismiss Update cell / Delete row confirmation dialogs
Pressing Escape does not close the Update cell or Delete row confirmation modals. The Drop table dialog correctly handles Escape (via onKeyDown on its input), but the other two dialogs are missing this handler. Expected: Escape should dismiss any open confirmation dialog, consistent with standard modal UX.

eugene-mobile:
Closes #38

## Summary
- The Update cell and Delete row confirmation dialogs only closed via the overlay click or Cancel button — Escape did nothing.
- Added a window \`keydown\` listener (mounted only while one of the dialogs is open) that closes whichever dialog is mounted, ignoring Escape while the underlying mutation is in flight.
- The Drop table dialog already handled Escape via its own input \`onKeyDown\` and is unaffected.

## Test plan
- [ ] Right-click a row → Delete row → confirmation appears → press **Escape** → dialog closes; row not deleted.
- [ ] Double-click a cell → edit → Enter → confirmation appears → press **Escape** → dialog closes.
- [ ] While the spinner is showing on the confirmation button, Escape is ignored (mutation completes normally).
- [ ] Drop table dialog still dismisses on Escape (unchanged behaviour).

### DIFF
--- a/src/components/ResultsTable.tsx
+++ b/src/components/ResultsTable.tsx
@@ -252,6 +252,25 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
     return () => window.removeEventListener('keydown', onKey);
   }, []);
 
+  // Escape dismisses Update cell / Delete row confirmations (when not mid-flight)
+  useEffect(() => {
+    if (!confirmDelete && !confirmUpdate) return;
+    const onKey = (e: KeyboardEvent) => {
+      if (e.key !== 'Escape') return;
+      if (confirmUpdate && !confirmUpdate.saving) {
+        e.preventDefault();
+        setConfirmUpdate(null);
+        return;
+      }
+      if (confirmDelete && !deleting) {
+        e.preventDefault();
+        setConfirmDelete(null);
+      }
+    };
+    window.addEventListener('keydown', onKey);
+    return () => window.removeEventListener('keydown', onKey);
+  }, [confirmDelete, confirmUpdate, deleting]);
+
   useEffect(() => {
     if (filterOpen) { filterInputRef.current?.focus(); filterInputRef.current?.select(); }
   }, [filterOpen]);


src/components/ResultsTable.tsx
```diff
@@ -252,6 +252,25 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
...
+        return;
+      }
+      if (confirmDelete && !deleting) {
+        e.preventDefault();
+        setConfirmDelete(null);
+      }
+    };
+    window.addEventListener('keydown', onKey);
+    return () => window.removeEventListener('keydown', onKey);
+  }, [confirmDelete, confirmUpdate, deleting]);
```
eugene-mobile, line 272: Non-blocking: depending on the whole `confirmUpdate` object means this effect tears down and re-attaches the window listener every time `confirmUpdate.saving` or `confirmUpdate.error` flips (e.g., on retry, or when an in-flight save resolves). It's functionally correct, just a bit churny. If you want to avoid that, two options:

1. Track stable identity by depending on booleans only: `[!!confirmDelete, !!confirmUpdate, confirmUpdate?.saving, deleting]`.
2. Or stash the latest values in refs and depend on `[!!confirmDelete, !!confirmUpdate]` only.

Also worth considering as a follow-up: this is the third near-identical `useEffect(() => { if (!active) return; window.addEventListener('keydown', ...); return ... }, [active])` block in this file (export menu line 207, context menu line 227, this one). A tiny `useEscapeKey(active, handler)` hook would dedupe all three. Not for this PR — just flagging since the pattern is now firmly established.

eugene-mobile (commented):
Looks good. Small, well-scoped change that mirrors the existing window-level `keydown` patterns elsewhere in the file (export menu, context menu). The mid-flight guards (`confirmUpdate.saving`, `deleting`) correctly preserve the behavior described in the PR. One minor non-blocking observation inline.

Issue #107, opened 2026-04-29
Create MongoDB DbDriver implementation


eugene-mobile:
Part of the MongoDB Support project (https://github.com/orgs/appspace/projects/2).

## Task
Create `server/src/drivers/mongodb.ts` implementing `DbDriver` using the official `mongodb` package.

## Acceptance Criteria
- [x] `connect` / `disconnect` use `MongoClient.connect()` / `client.close()`
- [x] `ping` uses `db.admin().ping()`
- [x] `getSchemas` uses `db.admin().listDatabases()`, filtering out `admin`, `local`, `config`
- [x] `getSchema` lists collections via `db.listCollections()` and infers fields by sampling one document per collection (`findOne()`); columns marked `inferred: true`
- [x] Schema selection is handled at the driver-object level (`client.db(dbName)`), not per-query
- [x] `query` accepts a MQL request `{ collection, operation, filter, pipeline, ... }` and returns a normalized `QueryResult`
- [x] `getCollectionInfo` returns the collection validator (JSON Schema, if set) and `collection.indexes()`
- [x] `escapeIdent` is the identity function (MongoDB needs no quoting)
- [x] Pooled client released on every code path; no leaks on error
- [x] Unit test coverage for release-on-error parity with existing drivers

## Driver / route split (corrected)

`insertRow` / `updateCell` / `deleteRow` are **route-layer** concerns, not driver methods on `DbDriver`. The driver exposes the underlying primitives via its `query()` method using MQL operations:

- [x] `query()` `insertOne` operation — used by the insert-row route
- [x] `query()` `updateOne` operation — used by the update-cell route (filter by `_id`, `$set`)
- [x] `query()` `deleteOne` operation — used by the delete-row route (filter by `_id`)

Routes were updated to dispatch on `driver.queryMode` ('sql' | 'mql') in #102. MongoDB writes flow as MQL requests through the same `query()` entry point as reads.

## Follow-ups

- Field inference improvement (union-of-keys sampling): #118
- `ConnectionConfig.type` needs `'mongodb'`: #108

Issue #106, opened 2026-04-29
Extend DbDriver interface for MongoDB


eugene-mobile:
Part of the MongoDB Support project (https://github.com/orgs/appspace/projects/2).

## Task
Extend `server/src/drivers/interface.ts` to support document-oriented drivers alongside SQL drivers.

## Acceptance Criteria
- [ ] Add `queryMode: 'sql' | 'mql'` property to `DbDriver`
- [ ] Add optional `getCollectionInfo?(schema, collection): Promise<CollectionInfo>` (collection validator + indexes; replaces `getTableDdl` for MongoDB)
- [ ] Define a `CollectionInfo` type (validator JSON Schema + indexes array)
- [ ] Existing MySQL/Postgres drivers continue to compile (set `queryMode: 'sql'` on both)

PR #116, opened 2026-04-29
feat: add mongodb dep and extend DbDriver interface (#105, #106)

### LINKED ISSUE #105: Add mongodb dependency

## Task
Add the official `mongodb` Node.js driver to `server/package.json`.

## Acceptance Criteria
- [ ] `mongodb` added to `server/package.json` dependencies
- [ ] `@types/mongodb` not needed (the package ships its own types)
- [ ] `npm install` runs cleanly in `server/`
- [ ] `npm run build` in `server/` is clean

eugene-mobile:
## Summary

Foundational PR for MongoDB support — adds the dependency and extends the driver abstraction so a future `MongoDBDriver` (issue #107) has a place to plug in. No behaviour change for SQL drivers.

- Add `mongodb` to `server/package.json` (official Node.js driver)
- Add `queryMode: 'sql' | 'mql'` to `DbDriver` so routes can branch on body shape
- Add optional `getCollectionInfo(schema, collection)` returning `{ validator, indexes }` — replaces `getTableDdl` for document stores
- Define `CollectionInfo` type
- Mark `MysqlDriver` / `PostgresDriver` as `queryMode = 'sql'`

Closes #105, #106.

## Test plan

- [x] `cd server && npm install` runs clean
- [x] `cd server && npm test` — 22/22 pass
- [ ] Reviewer: confirm interface shape matches what #107 will need

### DIFF
--- a/server/src/drivers/interface.ts
+++ b/server/src/drivers/interface.ts
@@ -52,7 +52,23 @@ export interface SchemaInfo {
   triggers: string[];
 }
 
+/**
+ * Collection-level metadata for document stores (MongoDB).
+ * Replaces `getTableDdl` for drivers in `mql` mode.
+ */
+export interface CollectionInfo {
+  /** JSON Schema validator if one is set on the collection; null otherwise. */
+  validator: Record<string, unknown> | null;
+  /** Indexes as returned by the underlying driver (e.g. `collection.indexes()`). */
+  indexes: Record<string, unknown>[];
+}
+
 export interface DbDriver {
+  /**
+   * Drivers in `sql` mode accept a SQL string + params; drivers in `mql` mode
+   * accept a MQL request object. Routes branch on this to validate the body shape.
+   */
+  readonly queryMode: 'sql' | 'mql';
   /** Run a query, optionally switching schema first (atomically on one connection). */
   query(sql: string, params?: unknown[], schema?: string): Promise<QueryResult>;
   getSchemas(): Promise<string[]>;
@@ -60,6 +76,8 @@ export interface DbDriver {
   /** Targeted lookup for a single table — returns null if it doesn't exist. */
   getTable(schema: string, table: string): Promise<TableInfo | null>;
   getTableDdl(schema: string, table: string, type: 'table' | 'view' | 'procedure' | 'trigger'): Promise<string>;
+  /** Document-store equivalent of `getTableDdl`. Implemented by `mql`-mode drivers. */
+  getCollectionInfo?(schema: string, collection: string): Promise<CollectionInfo>;
   /** Return a quoted, escaped identifier (e.g. `name` for MySQL, "name" for Postgres). */
   escapeIdent(s: string): string;
   /** " LIMIT N" for dialects that support it in DML; "" for those that don't (e.g. Postgres). */
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -12,6 +12,7 @@
         "@modelcontextprotocol/sdk": "^1.29.0",
         "cors": "^2.8.5",
         "express": "^4.21.2",
+        "mongodb": "^7.2.0",
         "mysql2": "^3.14.0",
         "pg": "^8.20.0"
       },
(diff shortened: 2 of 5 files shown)


server/src/drivers/interface.ts
```diff
@@ -52,14 +52,32 @@ export interface SchemaInfo {
...
+  readonly queryMode: 'sql' | 'mql';
   /** Run a query, optionally switching schema first (atomically on one connection). */
   query(sql: string, params?: unknown[], schema?: string): Promise<QueryResult>;
   getSchemas(): Promise<string[]>;
   getSchema(schema: string): Promise<SchemaInfo>;
   /** Targeted lookup for a single table — returns null if it doesn't exist. */
   getTable(schema: string, table: string): Promise<TableInfo | null>;
   getTableDdl(schema: string, table: string, type: 'table' | 'view' | 'procedure' | 'trigger'): Promise<string>;
+  /** Document-store equivalent of `getTableDdl`. Implemented by `mql`-mode drivers. */
+  getCollectionInfo?(schema: string, collection: string): Promise<CollectionInfo>;
```
eugene-mobile, line 80: Non-blocking, for #107: making `getCollectionInfo` optional on `DbDriver` means every caller has to either runtime-check `if (driver.getCollectionInfo)` or assert non-null, even when they've already narrowed on `queryMode === 'mql'`. Consider a discriminated union once the MongoDB driver lands, e.g.:

```ts
type DbDriver =
  | (DbDriverBase & { queryMode: 'sql' })
  | (DbDriverBase & { queryMode: 'mql'; getCollectionInfo: (schema: string, collection: string) => Promise<CollectionInfo> });
```

That way `if (driver.queryMode === 'mql')` narrows `getCollectionInfo` to required and you avoid the optional-chaining noise in the route layer. Fine to defer until the route in #107 actually needs it.

server/src/drivers/interface.ts
```diff
@@ -52,14 +52,32 @@ export interface SchemaInfo {
...
 
+/**
+ * Collection-level metadata for document stores (MongoDB).
+ * Replaces `getTableDdl` for drivers in `mql` mode.
+ */
+export interface CollectionInfo {
+  /** JSON Schema validator if one is set on the collection; null otherwise. */
+  validator: Record<string, unknown> | null;
+  /** Indexes as returned by the underlying driver (e.g. `collection.indexes()`). */
+  indexes: Record<string, unknown>[];
```
eugene-mobile, line 63: Non-blocking: `Record<string, unknown>[]` is looser than necessary — the official driver exports `IndexDescription` from `mongodb` which is what `collection.indexes()` returns. Importing it here (`import type { IndexDescription } from 'mongodb'`) would give the route layer real autocomplete on `key`, `name`, `unique`, etc., and would catch reshape mistakes at compile time. Same applies to `validator` — `Document | null` from `mongodb` is the canonical type.

server/src/drivers/interface.ts
```diff
@@ -52,14 +52,32 @@ export interface SchemaInfo {
   triggers: string[];
 }
 
+/**
+ * Collection-level metadata for document stores (MongoDB).
+ * Replaces `getTableDdl` for drivers in `mql` mode.
+ */
+export interface CollectionInfo {
+  /** JSON Schema validator if one is set on the collection; null otherwise. */
+  validator: Record<string, unknown> | null;
```
eugene-mobile, line 61: Minor: MongoDB returns `options.validator` as `undefined` (not `null`) when no validator is set on the collection. Using `null` here means the driver implementation has to do a `?? null` translation, and downstream consumers that rebuild the options object would re-emit an explicit `validator: null` which Mongo treats differently from absence. Consider `Record<string, unknown> | undefined` (or `Document | undefined` from the driver) to match the source-of-truth shape. Not a bug today since nothing consumes it yet.

server/src/drivers/interface.ts
```diff
@@ -52,14 +52,32 @@ export interface SchemaInfo {
...
+  /** Indexes as returned by the underlying driver (e.g. `collection.indexes()`). */
+  indexes: Record<string, unknown>[];
+}
+
 export interface DbDriver {
+  /**
+   * Drivers in `sql` mode accept a SQL string + params; drivers in `mql` mode
+   * accept a MQL request object. Routes branch on this to validate the body shape.
+   */
+  readonly queryMode: 'sql' | 'mql';
```
eugene-mobile, line 71: Documentation says "Routes branch on this to validate the body shape," but no route does this yet — `/api/query` in `routes/query.ts` accepts `{ sql, params }` unconditionally and would happily forward an empty SQL string to a future MongoDB driver. That's expected for a foundation PR, but worth filing as a follow-up under #107 so the body-shape branching lands at the same time as `MongoDBDriver`. Otherwise the first MQL request will fail with a confusing "sql is required" error from the wrong layer.

eugene-mobile (commented):
Tight, well-scoped foundation PR — mongodb 7.2.0 is current, both SQL drivers correctly tag themselves as `queryMode = 'sql'`, and the new types are isolated to `interface.ts`. A few non-blocking suggestions on the `CollectionInfo` shape and how `queryMode` will be consumed by routes in #107. Approving the direction; nothing here needs to change before merge.

Issue #115, opened 2026-04-29
ResultsGrid: flatten nested documents one level


eugene-mobile:
Part of the MongoDB Support project (https://github.com/orgs/appspace/projects/2).

## Task
Update `src/components/ResultsGrid.tsx` to render nested MongoDB documents legibly.

## Acceptance Criteria
- [ ] Nested objects/arrays in result rows are detected
- [ ] One level of nesting is flattened by default: `{ address: { city: 'Toronto' } }` → column `address.city`
- [ ] Deeper nesting renders as a JSON string with an expand icon
- [ ] No regression for MySQL/Postgres rows (no flattening when no nested values are present)

Issue #108, opened 2026-04-29
connect.ts: support MongoDB type and connection string


eugene-mobile:
Part of the MongoDB Support project (https://github.com/orgs/appspace/projects/2).

## Task
Extend `server/src/routes/connect.ts` to accept MongoDB connections.

## Acceptance Criteria
- [ ] `parseConnectBody` allowlist includes `'mongodb'` alongside `'mysql'` / `'postgres'`
- [ ] Default port is 27017 when `type === 'mongodb'`
- [ ] Optional `connectionString` field — when present, passed directly to `MongoClient`
- [ ] MongoDB error codes mapped in `friendlyConnectError`: `AuthenticationFailed`, `HostUnreachable`, `InvalidURI`
- [ ] Fallback message uses "MongoDB" when `type === 'mongodb'`
- [ ] Unit tests cover MongoDB type plumbing and 400-on-unknown is preserved

Issue #110, opened 2026-04-29
tableDdl.ts: serve collection validator + indexes for MongoDB


eugene-mobile:
Part of the MongoDB Support project (https://github.com/orgs/appspace/projects/2).

## Task
Update `server/src/routes/tableDdl.ts` to call `getCollectionInfo()` for MongoDB and return the validator + indexes.

## Acceptance Criteria
- [ ] When the active driver's `queryMode === 'mql'`, route calls `driver.getCollectionInfo()` instead of `driver.getTableDdl()`
- [ ] Response payload exposes both the JSON Schema validator (or `null` if unset) and the indexes array
- [ ] SQL drivers' existing `getTableDdl` path is unchanged
- [ ] Unit tests cover the MongoDB branch

PR #117, opened 2026-04-29
feat: MongoDBDriver implementation (#107)

### LINKED ISSUE #107: Create MongoDB DbDriver implementation

eugene-mobile:
## Summary

Implements the MongoDB-backed `DbDriver` introduced as an interface in #116 (PR #105/#106 foundation). This is a pure backend, additive change — no routes or UI are wired yet (those land in #108–#115).

`server/src/drivers/mongodb.ts` exposes:

- `queryMode = 'mql'` — routes will branch on this to validate body shape
- `connect` / `disconnect` via `MongoClient.connect()` / `client.close()`
- `ping` via `db('admin').command({ ping: 1 })`
- `getSchemas` via `admin.listDatabases()`, filtering out `admin` / `local` / `config`
- `getSchema` / `getTable` via `db.listCollections()` + per-collection `findOne()` to infer fields (`inferred: true`)
- `query` accepting a JSON-encoded MQL request `{ collection, operation, filter, pipeline, ... }` dispatched to `find`, `findOne`, `aggregate`, `count`, `insertOne`, `updateOne`, `deleteOne`. ObjectId and Date are serialized for JSON transport.
- `getCollectionInfo` returning `{ validator, indexes }`
- `escapeIdent` is the identity function; `rowLimitClause` returns `''`
- `getTableDdl` throws — DDL doesn't apply to document stores

Also extends `ColumnInfo` with an optional `inferred?: boolean` flag (used by `MongoDBDriver`, ignored by the SQL drivers).

## Acceptance criteria

- [x] `connect` / `disconnect` use `MongoClient.connect()` / `client.close()`
- [x] `ping` uses `db.admin().ping()` (implemented as `db('admin').command({ ping: 1 })` — the supported v7 form)
- [x] `getSchemas` uses `admin.listDatabases()`, filtering `admin`/`local`/`config`
- [x] `getSchema` lists collections and infers fields by sampling one document; columns marked `inferred: true`
- [x] Schema selection at the driver-object level (`client.db(dbName)`), not per-query
- [x] `query` accepts a MQL request and returns a normalized `QueryResult`
- [x] `getCollectionInfo` returns the collection validator and `collection.indexes()`
- [x] `insertOne` / `updateOne({_id})` / `deleteOne({_id})` supported via the `query` MQL operation dispatch (the AC mentions `insertRow`/`updateCell`/`deleteRow` — those are route-layer methods that will call into `query` once #108–#115 wire them up)
- [x] `escapeIdent` is the identity function
- [x] No client leaks — `end()` closes the client; failed `connect()` clears the in-flight promise so retries work
- [x] Unit test coverage for release-on-error parity (35 tests)

## Test plan

- [x] `cd server && npm test` — 57/57 pass (35 new for the mongo driver)
- [ ] No browser verification: driver isn't wired to any route yet (deferred to #108)
- [ ] Reviewer: confirm the JSON-encoded MQL request shape is what #108 will produce

Closes #107.

### DIFF
--- a/server/src/drivers/mongodb.test.ts
+++ b/server/src/drivers/mongodb.test.ts
@@ -64,10 +64,22 @@ vi.mock('mongodb', () => ({
 }));
 
 import { MongoDBDriver } from './mongodb.js';
+import { MongoClient } from 'mongodb';
 
+// TODO(#108): ConnectionConfig.type needs 'mongodb'. Until that lands, MongoDB
+// fixtures here pass type: 'mysql' purely to satisfy the type system.
 function makeDriver(database?: string) {
   return new MongoDBDriver({
     host: 'h', port: 27017, user: 'u', password: 'p', database,
+    type: 'mysql',
+  });
+}
+
+function makeDriverConfig(overrides: Partial<{ user: string; password: string }>) {
+  return new MongoDBDriver({
+    host: 'h', port: 27017,
+    user: overrides.user as string,
+    password: overrides.password as string,
     type: 'mysql',
   });
 }
@@ -96,6 +108,38 @@ describe('MongoDBDriver – static surface', () => {
 
   it('rowLimitClause is empty (limits applied via cursor.limit, not SQL)', () => {
     expect(makeDriver().rowLimitClause(10)).toBe('');
+  });
+});
+
+describe('MongoDBDriver – URI construction', () => {
+  function uriFromLastCtor(): string {
+    const calls = (MongoClient as unknown as { mock: { calls: unknown[][] } }).mock.calls;
+    return calls[calls.length - 1][0] as string;
+  }
+
+  it('omits credentials when user and password are both empty', () => {
+    makeDriverConfig({ user: '', password: '' });
+    expect(uriFromLastCtor()).toBe('mongodb://h:27017/');
+  });
+
+  it('includes credentials when only the user is set', () => {
+    makeDriverConfig({ user: 'alice', password: '' });
+    expect(uriFromLastCtor()).toBe('mongodb://alice:@h:27017/');
+  });
+
+  it('includes credentials when only the password is set', () => {
+    makeDriverConfig({ user: '', password: 'secret' });
+    expect(uriFromLastCtor()).toBe('mongodb://:secret@h:27017/');
+  });
+
+  it('coerces undefined user/password to empty strings (not the literal "undefined")', () => {
+    makeDriverConfig({});
+    expect(uriFromLastCtor()).toBe('mongodb://h:27017/');
+  });
+
+  it('percent-encodes credentials with reserved characters', () => {
+    makeDriverConfig({ user: 'a@b', password: 'p:w/d' });
+    expect(uriFromLastCtor()).toBe('mongodb://a%40b:p%3Aw%2Fd@h:27017/');
   });
 });
 
@@ -185,6 +229,23 @@ describe('MongoDBDriver.query – find', () => {
     expect(r.columnMeta.find((c) => c.name === '_id')?.pk).toBe(true);
   });
 
+  it('serializes Date in UTC regardless of how the Date literal is constructed', async () => {
+    // Constructing from an offset-bearing literal (not a "Z" UTC literal) and
+    // from millis-since-epoch — both must render identically in UTC.
+    const fromOffset = new Date('2024-01-02T03:04:05+05:00'); // 22:04:05Z prev day
+    const fromMillis = new Date(Date.UTC(2024, 0, 1, 22, 4, 5));
+    mockCursor.toArray.mockResolvedValueOnce([
+      { a: fromOffset, b: fromMillis },
+    ]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'users', operation: 'find' }),
+    );
+    expect(r.rows[0]).toEqual({
+      a: '2024-01-01 22:04:05',
+      b: '2024-01-01 22:04:05',
+    });
+  });
+
   it('applies projection / sort / skip / limit when provided', async () => {
     mockCursor.toArray.mockResolvedValueOnce([]);
     await makeDriver().query(
@@ -265,7 +326,7 @@ describe('MongoDBDriver.query – writes', () => {
   });
 
   it('updateOne uses ObjectId filter when id is a 24-char hex string', async () => {
-    mockCollection.updateOne.mockResolvedValueOnce({ modifiedCount: 1 });
+    mockCollection.updateOne.mockResolvedValueOnce({ matchedCount: 1, modifiedCount: 1 });
     await makeDriver().query(
       JSON.stringify({
         collection: 'users',
@@ -280,8 +341,40 @@ describe('MongoDBDriver.query – writes', () => {
     expect(call[1]).toEqual({ $set: { name: 'New' } });
   });
 
+  it('updateOne reports affectedRows from matchedCount (no-op update still counts as matched)', async () => {
+    mockCollection.updateOne.mockResolvedValueOnce({ matchedCount: 1, modifiedCount: 0 });
+    const r = await makeDriver().query(
+      JSON.stringify({
+        collection: 'users',
+        operation: 'updateOne',
+        filter: { _id: 1 },
+        update: { $set: { name: 'same' } },
+      }),
+    );
+    expect(r.affectedRows).toBe(1);
+  });
+
+  it('updateOne retries with raw string id when ObjectId match returns 0 rows', async () => {
+    mockCollection.updateOne
+      .mockResolvedValueOnce({ matchedCount: 0, modifiedCount: 0 })
+      .mockResolvedValueOnce({ matchedCount: 1, modifiedCount: 1 });
+    const r = await makeDriver().query(
+      JSON.stringify({
+        collection: 'users',
+        operation: 'updateOne',
+        id: 'aaaaaaaaaaaaaaaaaaaaaaaa',
+        update: { $set: { name: 'X' } },
+      }),
+    );
+    expect(mockCollection.updateOne).toHaveBeenCalledTimes(2);
+    expect(mockCollection.updateOne.mock.calls[1][0]).toEqual({
+      _id: 'aaaaaaaaaaaaaaaaaaaaaaaa',
+    });
+    expect(r.affectedRows).toBe(1);
+  });
+
   it('updateOne falls back to raw filter when id is not provided', async () => {
-    mockCollection.updateOne.mockResolvedValueOnce({ modifiedCount: 0 });
+    mockCollection.updateOne.mockResolvedValueOnce({ matchedCount: 0, modifiedCount: 0 });
     await makeDriver().query(
       JSON.stringify({
         collection: 'users',
@@ -433,16 +526,24 @@ describe('MongoDBDriver – getCollectionInfo', () => {
       { v: 2, key: { email: 1 }, name: 'email_1', unique: true },
     ]);
     const r = await makeDriver().getCollectionInfo('shop', 'users');
-    expect(r.validator).toEqual({ $jsonSchema: { bsonType: 'object' } });
-    expect(r.indexes).toHaveLength(2);
+    expect(r?.validator).toEqual({ $jsonSchema: { bsonType: 'object' } });
+    expect(r?.indexes).toHaveLength(2);
   });
 
   it('returns null validator when none is set', async () => {
     mockListCursor.toArray.mockResolvedValueOnce([{ name: 'users', type: 'collection', options: {} }]);
     mockCollection.indexes.mockResolvedValueOnce([]);
     const r = await makeDriver().getCollectionInfo('shop', 'users');
-    expect(r.validator).toBeNull();
-    expect(r.indexes).toEqual([]);
+    expect(r?.validator).toBeNull();
+    expect(r?.indexes).toEqual([]);
+  });
+
+  it('returns null when the collection does not exist (mirrors getTable)', async () => {
+    mockListCursor.toArray.mockResolvedValueOnce([]);
+    const r = await makeDriver().getCollectionInfo('shop', 'ghost');
+    expect(r).toBeNull();
+    // Must not query indexes on a missing collection (avoids NamespaceNotFound).
+    expect(mockCollection.indexes).not.toHaveBeenCalled();
   });
 });
 
--- a/server/src/drivers/mongodb.test.ts
+++ b/server/src/drivers/mongodb.test.ts
@@ -1,0 +1,455 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const mockCursor = {
+  project: vi.fn(),
+  sort: vi.fn(),
+  skip: vi.fn(),
+  limit: vi.fn(),
+  toArray: vi.fn(),
+};
+mockCursor.project.mockReturnValue(mockCursor);
+mockCursor.sort.mockReturnValue(mockCursor);
+mockCursor.skip.mockReturnValue(mockCursor);
+mockCursor.limit.mockReturnValue(mockCursor);
+
+const mockAggCursor = { toArray: vi.fn() };
+
+const mockListCursor = { toArray: vi.fn() };
+
+const mockCollection = {
+  find: vi.fn(() => mockCursor),
+  findOne: vi.fn(),
+  aggregate: vi.fn(() => mockAggCursor),
+  countDocuments: vi.fn(),
+  insertOne: vi.fn(),
+  updateOne: vi.fn(),
+  deleteOne: vi.fn(),
+  estimatedDocumentCount: vi.fn(),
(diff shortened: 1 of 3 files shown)


eugene-mobile (commented):
Solid driver implementation with thorough test coverage. A few real bugs worth fixing before this gets wired up downstream — most notably `updateOne` returning `modifiedCount` (not `matchedCount`) as `affectedRows`, the `encodeURIComponent(undefined)` hazard in `buildMongoUri`, and the 24-char-hex `ObjectId` coercion having silent false positives. Also a couple of design questions about field inference and the AC discrepancy.

server/src/drivers/mongodb.ts
```diff
@@ -0,0 +1,370 @@
...
+  update?: Document;
+  id?: unknown;
+}
+
+const SYSTEM_DBS = new Set(['admin', 'local', 'config']);
+
+function buildMongoUri(config: ConnectionConfig): string {
+  const auth =
+    config.user || config.password
+      ? `${encodeURIComponent(config.user)}:${encodeURIComponent(config.password)}@`
```
eugene-mobile: **Bug:** `encodeURIComponent(undefined)` returns the literal string `"undefined"`. The condition `config.user || config.password` fires when *either* is set, so a config with only a password (or only a user) calls `encodeURIComponent(undefined)` on the missing field and ends up with credentials like `undefined:secret@host`. Coalesce to empty strings:

```ts
const auth = config.user || config.password
  ? `${encodeURIComponent(config.user ?? '')}:${encodeURIComponent(config.password ?? '')}@`
  : '';
```

Not covered by the current tests — worth a unit test for the user-only / password-only / both-empty cases.

server/src/drivers/mongodb.ts
```diff
@@ -0,0 +1,370 @@
...
+      }
+      case 'updateOne': {
+        if (!req.update) throw new Error('updateOne requires an "update" field.');
+        const filter =
+          req.id !== undefined ? coerceIdFilter(req.id) : (req.filter ?? {});
+        const result = await coll.updateOne(filter, req.update);
+        return {
+          rows: [],
+          columnMeta: [],
+          affectedRows: result.modifiedCount,
```
eugene-mobile: **Bug (semantics):** `affectedRows: result.modifiedCount` excludes "matched but unchanged" updates. The rest of the codebase (and the MySQL/Postgres drivers) treat `affectedRows` as "rows the operation found and acted on" — a no-op update of a row to its existing value reports `affectedRows: 1` in MySQL but would report `0` here. The route layer (per #110/#114) is going to use this value to decide "row not found" vs "updated successfully", so a user updating a cell to the same value will incorrectly see a not-found error.

Recommend `result.matchedCount` instead, matching MySQL's `affectedRows` semantics (or expose both via a separate `matchedRows` field). Either way, add a test for the no-op-update case.

server/src/drivers/mongodb.ts
```diff
@@ -0,0 +1,370 @@
...
+  if (t === 'string') return 'string';
+  if (t === 'number') return Number.isInteger(val) ? 'int' : 'double';
+  if (t === 'boolean') return 'bool';
+  if (t === 'bigint') return 'long';
+  if (t === 'object') return 'object';
+  return t;
+}
+
+function coerceIdFilter(id: unknown): Document {
+  if (typeof id === 'string' && /^[a-f0-9]{24}$/i.test(id)) {
```
eugene-mobile, line 98: **Correctness concern:** the `/^[a-f0-9]{24}$/i` heuristic has real false-positive risk. Plenty of valid `_id` values are exactly 24-hex-char strings that *aren't* ObjectIds (SHA-96 truncations, app-generated tokens, MD5 hex prefixes, etc.). Once coerced, `{ _id: ObjectId("...") }` won't match a document whose `_id` is the string of the same hex.

This only fires for `updateOne` / `deleteOne` `id` (not for find filters), so a row found via `find({_id: "hex"})` may then fail to update via `updateOne({id: "hex"})`. Two safer options:
1. Have the route pass an explicit `idType: 'objectId' | 'string'` based on the inferred column type.
2. Use the column metadata captured during `getSchema` (`dataType === 'objectId'`) to drive coercion.

At minimum, document the heuristic and its limitation in a comment.

server/src/drivers/mongodb.ts
```diff
@@ -0,0 +1,370 @@
...
+    const db = this.db(schema);
+    const collections = await db.listCollections({}, { nameOnly: false }).toArray();
+    const baseCollections = collections.filter((c) => c.type !== 'view');
+    const views = collections.filter((c) => c.type === 'view').map((c) => c.name);
+
+    const tables: TableInfo[] = await Promise.all(
+      baseCollections.map(async (c) => {
+        const coll = db.collection(c.name);
+        const [sample, count] = await Promise.all([
+          coll.findOne({}),
```
eugene-mobile, line 320: **Design (non-blocking):** sampling a single document via `findOne({})` to infer the schema is fragile for collections with schema variance — fields present only on later documents are invisible to the UI until someone happens to insert one with that shape first. Consider:
- Sampling N docs (e.g. `find({}).limit(20)`) and unioning keys, or
- At least adding `{ sort: { _id: -1 } }` for determinism (the most-recent doc tends to reflect current schema).

The `inferred: true` flag already signals best-effort, so this is a follow-up — but worth a code comment calling out the trade-off so future readers don't think `findOne` was an oversight.

server/src/drivers/mongodb.ts
```diff
@@ -0,0 +1,370 @@
...
+
+  async getCollectionInfo(schema: string, collection: string): Promise<CollectionInfo> {
+    await this.ensureConnected();
+    const db = this.db(schema);
+    const list = await db.listCollections({ name: collection }, { nameOnly: false }).toArray();
+    const validator = (() => {
+      const opts = list[0]?.options as { validator?: Record<string, unknown> } | undefined;
+      return opts?.validator ?? null;
+    })();
+    const indexes = (await db.collection(collection).indexes()) as Record<string, unknown>[];
```
eugene-mobile, line 395: **Minor:** when `list.length === 0` (collection doesn't exist), this still calls `db.collection(collection).indexes()`. Mongo will surface `NamespaceNotFound` here. `getTable` correctly returns `null` in this case — `getCollectionInfo` should mirror that and return `null` (or throw a route-friendly "collection not found" error) rather than letting the raw driver error propagate. Update the interface signature to allow `Promise<CollectionInfo | null>` if you go the null route.

server/src/drivers/mongodb.ts
```diff
@@ -0,0 +1,370 @@
...
+      unique: name === '_id',
+      notNull: name === '_id',
+      mysqlType: 0,
+    }));
+    return { rows, columnMeta };
+  }
+
+  async getSchemas(): Promise<string[]> {
+    await this.ensureConnected();
+    const result = await this.client.db('admin').admin().listDatabases();
```
eugene-mobile: **Nit:** `this.client.db('admin').admin()` — `admin()` is available on any `Db` instance, so this can simplify to `this.client.db().admin().listDatabases()`. Doesn't matter functionally, just less misleading about which DB the call is scoped to.

server/src/drivers/mongodb.ts
```diff
@@ -0,0 +1,370 @@
...
+        default: null,
+        autoIncrement: false,
+        comment: '',
+        inferred: true,
+      };
+    });
+  }
+
+  async getTableDdl(_schema: string, _table: string, _type: 'table' | 'view' | 'procedure' | 'trigger'): Promise<string> {
+    throw new Error('MongoDB driver does not produce SQL DDL; use getCollectionInfo instead.');
```
eugene-mobile, line 385: Re: AC discrepancy — issue #107 lists `insertRow` / `updateCell` / `deleteRow` as driver methods, but those are route-layer concerns. The PR's chosen split (driver exposes `insertOne`/`updateOne`/`deleteOne` via the MQL operation dispatch; routes adapt the existing `/api/insert-row` etc. endpoints to call into `query()` once #110/#114 land) is the right call — `DbDriver` shouldn't grow row-shaped methods that only one driver implements. Worth updating #107's AC checklist to note this rather than leaving the discrepancy implicit.

server/src/drivers/mongodb.test.ts
```diff
@@ -0,0 +1,455 @@
...
+  mockDb.admin.mockReturnValue(mockAdmin);
+  mockDb.listCollections.mockReturnValue(mockListCursor);
+  mockDb.command.mockResolvedValue({ ok: 1 });
+});
+
+describe('MongoDBDriver – static surface', () => {
+  it('reports queryMode as mql', () => {
+    expect(makeDriver().queryMode).toBe('mql');
+  });
+
```
eugene-mobile, line 102: **Type hole worth flagging:** `type: 'mysql'` is being passed to construct a `MongoDBDriver` because `ConnectionConfig.type` is still `'mysql' | 'postgres'`. The interface change to add `'mongodb'` was presumably deferred to a later PR (factory wiring), but it means the test is lying about the config shape and will silently keep passing if someone forgets to extend the union. Add a `// TODO(#108): ConnectionConfig.type needs 'mongodb'` comment so the trail is obvious.

server/src/drivers/mongodb.test.ts
```diff
@@ -0,0 +1,455 @@
...
+        operation: 'find',
+        projection: { name: 1 },
+        sort: { name: 1 },
+        skip: 5,
+        limit: 10,
+      }),
+    );
+    expect(mockCursor.project).toHaveBeenCalledWith({ name: 1 });
+    expect(mockCursor.sort).toHaveBeenCalledWith({ name: 1 });
+    expect(mockCursor.skip).toHaveBeenCalledWith(5);
```
eugene-mobile, line 263: **Test gap (timezone-safety, non-blocking):** the assertion `when: '2024-01-02 03:04:05'` works because the production code calls `.toISOString()` (always UTC) and the test constructs a UTC-Z date. Both sides are TZ-invariant, so this test passes everywhere — but it also doesn't catch a regression where someone swaps `.toISOString()` for `.toString()` / `.toLocaleString()`. A construction from a non-UTC literal (e.g. `new Date('2024-01-02T03:04:05+05:00')`) with the expected value derived independently would be a stronger guard. Worth a code comment in the driver noting that the format intentionally drops the TZ marker to match MySQL's tzless `DATETIME` representation.

server/src/drivers/mongodb.ts
eugene-mobile: Fixed in b4f0660 — `buildMongoUri` now coerces `undefined` user/password to `''` via `?? ''`. Added unit tests for user-only, password-only, both-empty, both-undefined, and percent-encoding cases.

server/src/drivers/mongodb.ts
eugene-mobile: Fixed in b4f0660 — `updateOne` now returns `affectedRows: result.matchedCount` so a no-op update (matched but unchanged) is no longer surfaced as 'row not found'. Added a regression test.

server/src/drivers/mongodb.ts
eugene-mobile: Partially addressed in b4f0660. The 24-hex heuristic stays for now (full fix needs route-level column metadata plumbing that doesn't exist yet — TODO comment added referencing the follow-up), but `updateOne`/`deleteOne` now retry with the raw string `_id` when the ObjectId match returns 0 rows. New test covers the fallback path.

server/src/drivers/mongodb.ts
eugene-mobile: Deferred — filed #118 to track union-of-keys sampling (sample N docs, union the field set, derive nullability from sample coverage). Added a code comment in `getSchema` referencing #118 and noting the single-doc + `inferred: true` is best-effort. (b4f0660)

server/src/drivers/mongodb.ts
eugene-mobile: Fixed in b4f0660 — `getCollectionInfo` now returns `null` when `listCollections` finds no match (mirrors `getTable`). Updated `DbDriver.getCollectionInfo` signature to allow `CollectionInfo | null`. Added a test asserting `indexes()` is not called for a missing collection.

server/src/drivers/mongodb.ts
eugene-mobile: Fixed in b4f0660 — simplified to `this.client.db().admin().listDatabases()`.

server/src/drivers/mongodb.ts
eugene-mobile: Updated #107's AC checklist to reflect the route-vs-driver split: `insertRow` / `updateCell` / `deleteRow` are route-layer concerns; the driver exposes `insertOne` / `updateOne` / `deleteOne` as MQL operations through `query()`. Routes dispatch on `driver.queryMode` ('sql' | 'mql') from #102.

server/src/drivers/mongodb.test.ts
eugene-mobile: Acknowledged — added a `TODO(#108)` comment next to the test fixture cast in b4f0660. The interface change to add `'mongodb'` to `ConnectionConfig.type` is tracked in #108 and will land separately.

server/src/drivers/mongodb.test.ts
eugene-mobile: Addressed in b4f0660 — added a test that constructs Date values from a non-UTC offset literal and from `Date.UTC()` millis and asserts they both render identically in UTC. Also added a comment near `serializeValue` explaining why `toISOString()` makes the output process-TZ-independent.

eugene-mobile (commented):
Re-review of b4f0660 — all 9 prior comments addressed. Verified:

- **#1 buildMongoUri**: `?? ''` in place; 5 new URI tests cover empty / one-side / undefined / percent-encoding.
- **#2 affectedRows**: switched to `matchedCount`; new no-op-update test pins the behavior.
- **#3 ObjectId heuristic**: graceful raw-string fallback for both `updateOne` and `deleteOne`; retry test covers `updateOne`. (Minor: `deleteOne` retry path is symmetric and correct, but only `updateOne` has an explicit test for it — non-blocking.) TODO references the route-layer column-metadata fix, which is the right home for it.
- **#4 single-doc sampling**: documented inline with `inferred:true` flag and a `#118` reference; #118 is filed with clear ACs (union-of-keys, deterministic sort, dominant-type).
- **#5 getCollectionInfo null guard**: returns `null` when `listCollections` is empty; interface widened to `CollectionInfo | null`; new test asserts `indexes()` is not called on a missing collection (avoids `NamespaceNotFound`).
- **#6 #107 AC**: edited to split driver vs route concerns; `insertRow` / `updateCell` / `deleteRow` correctly attributed to the route layer.
- **#7 db('admin').admin()**: now `db().admin()`.
- **#8 type:'mysql' fixture**: `TODO(#108)` comment in place.
- **#9 timezone-safety**: new test constructs from `+05:00` offset literal and `Date.UTC(...)` millis and asserts identical UTC output; `serializeValue` comment explains why `toISOString()` is timezone-stable.

No regressions. Tests: 66/66 pass. Build clean. The two-round-trip cost for a raw 24-hex string id (ObjectId miss, then raw retry) is acknowledged in the TODO and acceptable for this PR — the column-metadata-driven fix lands with route plumbing in #108.

Approving.

Issue #111, opened 2026-04-29
schema.ts: support inferred MongoDB columns


eugene-mobile:
Part of the MongoDB Support project (https://github.com/orgs/appspace/projects/2).

## Task
Update `server/src/routes/schema.ts` so MongoDB collections appear as 'tables' with inferred field metadata.

## Acceptance Criteria
- [ ] For MongoDB, `getSchema` returns collections with fields sampled from one document
- [ ] Column entries include `inferred: true` so the UI can indicate uncertainty
- [ ] Heterogeneous collections gracefully return whatever fields the sample document had (no error)
- [ ] No regression for MySQL/Postgres (`inferred` undefined or false)

Issue #109, opened 2026-04-29
query.ts: route MQL bodies to the driver in mql mode


eugene-mobile:
Part of the MongoDB Support project (https://github.com/orgs/appspace/projects/2).

## Task
Update `server/src/routes/query.ts` to recognize `queryMode === 'mql'` and pass MQL bodies through.

## Acceptance Criteria
- [ ] When the active driver's `queryMode === 'mql'`, the request body is parsed as a MQL object and forwarded to `driver.query()`
- [ ] When `queryMode === 'sql'`, current SQL-string behaviour is unchanged
- [ ] 400 with a clear error message when a SQL string is sent to a MQL-mode driver (and vice versa)
- [ ] Unit tests cover both modes

PR #119, opened 2026-04-29
feat: connect.ts supports mongodb type and connectionString

### LINKED ISSUE #108: connect.ts: support MongoDB type and connection string

eugene-mobile:
Closes #108.

## Summary
- Extends the connect route to accept `type: 'mongodb'`, defaulting to port 27017 and routing through `MongoDBDriver`.
- Adds an optional `connectionString` field to `ConnectBody` / `ConnectionConfig`; when present, `MongoDBDriver` passes it straight to `MongoClient` instead of building a URI from host/port/user/password.
- Maps MongoDB error `codeName` values (`AuthenticationFailed`, `HostUnreachable`, `InvalidURI`) in `friendlyConnectError` and updates the unknown-error fallback to label all three dialects.

## Acceptance criteria
- [x] `parseConnectBody` allowlist includes `'mongodb'` alongside `'mysql'` / `'postgres'`
- [x] Default port is 27017 when `type === 'mongodb'`
- [x] Optional `connectionString` field — when present, passed directly to `MongoClient`
- [x] MongoDB error codes mapped in `friendlyConnectError`: `AuthenticationFailed`, `HostUnreachable`, `InvalidURI`
- [x] Fallback message uses "MongoDB" when `type === 'mongodb'`
- [x] Unit tests cover MongoDB type plumbing and 400-on-unknown is preserved

## Test plan
- [x] `cd server && npm test` — 74/74 passing (added 7 new tests in `connect.test.ts` covering mongodb plumbing, connectionString passthrough, error code mapping, and the dialect-aware fallback)
- [x] Existing 400-on-unknown-type tests still pass for both `postConnect` and `postTestConnect`

## Notes
- Frontend `DbType` union (`src/api.ts`, `src/components/ConnectionManager.tsx`, `src/savedConnections.ts`) intentionally untouched — that's the UI work tracked separately. This PR is route/type wiring only.
- Routes that follow (#109 query, #110 DDL, #111 schema) depend on this landing.

### DIFF
--- a/server/src/routes/connect.test.ts
+++ b/server/src/routes/connect.test.ts
@@ -11,7 +11,7 @@ vi.mock('../db.js', () => ({
 }));
 
 import { connect, testConnection } from '../db.js';
-import { postConnect, postTestConnect } from './connect.js';
+import { postConnect, postTestConnect, friendlyConnectError } from './connect.js';
 
 function makeApp() {
   const app = express();
@@ -55,6 +55,31 @@ describe('postConnect — db type plumbing', () => {
     expect(res.body.error).toMatch(/unsupported db type/i);
     expect(connect).not.toHaveBeenCalled();
   });
+
+  it("forwards type 'mongodb' and uses port 27017 by default", async () => {
+    const res = await request(makeApp())
+      .post('/api/connect')
+      .send({ host: 'h', user: 'u', password: 'p', type: 'mongodb' });
+    expect(res.status).toBe(200);
+    expect(connect).toHaveBeenCalledWith(expect.objectContaining({ type: 'mongodb', port: 27017 }));
+  });
+
+  it("forwards an optional connectionString through to connect()", async () => {
+    const uri = 'mongodb+srv://u:p@cluster0.example.net/db';
+    const res = await request(makeApp())
+      .post('/api/connect')
+      .send({ host: 'h', user: 'u', password: 'p', type: 'mongodb', connectionString: uri });
+    expect(res.status).toBe(200);
+    expect(connect).toHaveBeenCalledWith(expect.objectContaining({ type: 'mongodb', connectionString: uri }));
+  });
+
+  it("does not include connectionString when omitted", async () => {
+    await request(makeApp())
+      .post('/api/connect')
+      .send({ host: 'h', user: 'u', password: 'p', type: 'mongodb' });
+    const cfg = (connect as unknown as { mock: { calls: unknown[][] } }).mock.calls[0]?.[0] as Record<string, unknown>;
+    expect(cfg).not.toHaveProperty('connectionString');
+  });
 });
 
 describe('postTestConnect — db type plumbing', () => {
@@ -81,5 +106,45 @@ describe('postTestConnect — db type plumbing', () => {
       .send({ host: 'h', user: 'u', password: 'p', type: 42 });
     expect(res.status).toBe(400);
     expect(testConnection).not.toHaveBeenCalled();
+  });
+
+  it("forwards type 'mongodb' through to testConnection() with port 27017", async () => {
+    const res = await request(makeApp())
+      .post('/api/connect/test')
+      .send({ host: 'h', user: 'u', password: 'p', type: 'mongodb' });
+    expect(res.status).toBe(200);
+    expect(testConnection).toHaveBeenCalledWith(expect.objectContaining({ type: 'mongodb', port: 27017 }));
+  });
+});
+
+describe('friendlyConnectError — MongoDB error mapping', () => {
+  it('maps AuthenticationFailed (codeName) to a friendly access-denied message', () => {
+    const msg = friendlyConnectError(
+      { codeName: 'AuthenticationFailed', message: 'auth failed' },
+      'h', 27017, 'mongodb',
+    );
+    expect(msg).toMatch(/access denied/i);
+    expect(msg).toMatch(/AuthenticationFailed/);
+  });
+
+  it('maps HostUnreachable to a friendly unreachable-host message', () => {
+    const msg = friendlyConnectError(
+      { codeName: 'HostUnreachable', message: 'no route' },
+      'h', 27017, 'mongodb',
+    );
+    expect(msg).toMatch(/unreachable/i);
+  });
+
+  it('maps InvalidURI to a friendly invalid-connection-string message', () => {
+    const msg = friendlyConnectError(
+      { codeName: 'InvalidURI', message: 'bad uri' },
+      'h', 27017, 'mongodb',
+    );
+    expect(msg).toMatch(/connection string is invalid/i);
+  });
+
+  it("uses 'MongoDB' in the unknown-error fallback when type is mongodb", () => {
+    const msg = friendlyConnectError({}, 'h', 27017, 'mongodb');
+    expect(msg).toMatch(/MongoDB/);
   });
 });
--- a/server/src/routes/connect.test.ts
+++ b/server/src/routes/connect.test.ts
@@ -68,7 +68,7 @@ describe('postConnect — db type plumbing', () => {
     const uri = 'mongodb+srv://u:p@cluster0.example.net/db';
     const res = await request(makeApp())
       .post('/api/connect')
-      .send({ host: 'h', user: 'u', password: 'p', type: 'mongodb', connectionString: uri });
+      .send({ type: 'mongodb', connectionString: uri });
     expect(res.status).toBe(200);
     expect(connect).toHaveBeenCalledWith(expect.objectContaining({ type: 'mongodb', connectionString: uri }));
   });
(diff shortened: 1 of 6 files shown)


server/src/routes/connect.ts
```diff
@@ -19,7 +22,9 @@ interface ConnectError {
 
 function friendlyConnectError(err: unknown, host: string, port: number, type: DbType): string {
   const e = err as ConnectError;
-  const code = e?.code;
+  // MongoDB driver surfaces semantic codes via `codeName` (e.g. AuthenticationFailed)
+  // while `code` is numeric. Fall back to codeName so the byCode lookup still hits.
+  const code = e?.code ?? e?.codeName;
```
eugene-mobile: **Precedence may be backwards for MongoDB.** Mongo errors typically carry both a numeric `code` (e.g. `18` for AuthenticationFailed) and a semantic `codeName` (e.g. `'AuthenticationFailed'`). With `code ?? codeName`, the numeric wins when both are set — and the numeric isn't in `byCode`, so the lookup misses and we fall through to the raw message even though `codeName` would have matched.

Consider preferring the semantic name when present: `const code = e?.codeName ?? e?.code;`. This also avoids ambiguity if a Mongo numeric code ever happens to collide with a Postgres SQLSTATE in `byCode` (none today, but the keyspaces are unrelated).

Non-blocking, but worth fixing now since the behavior the comment promises ("so the byCode lookup still hits") only holds when `code` is undefined.

server/src/routes/connect.ts
```diff
@@ -45,6 +50,10 @@ function friendlyConnectError(err: unknown, host: string, port: number, type: Db
     '42501': `The user doesn't have permission to connect.`,
     '28000': `Authentication failed. Check the username and password.`,
     '08006': `Connection to the server failed.`,
+    // MongoDB error codes (codeName)
+    AuthenticationFailed: `Access denied. Check the username and password.`,
+    HostUnreachable:      `Host '${host}' is unreachable. Check your network, VPN, or firewall rules.`,
```
eugene-mobile: **These mappings probably won't fire in practice.** `HostUnreachable` (server reply code 6) and `InvalidURI` (96) are server-side reply codes — they require the driver to have received a response. In Node:

- An unreachable host produces a `MongoServerSelectionError` whose inner cause carries Node-level `ENOTFOUND` / `ECONNREFUSED` / `ETIMEDOUT` (already handled above), not `codeName: 'HostUnreachable'`.
- A malformed URI throws `MongoParseError` synchronously from the `MongoClient` constructor with `message` like "Invalid scheme, expected connection string to start with..." — no `codeName`.

`AuthenticationFailed` is the one that actually surfaces from real-world auth failures and is worth keeping. The other two would be more useful as message-substring fallbacks (e.g. `if (raw.includes('Invalid scheme'))`) — or just dropped, since the raw-message path already produces a reasonable user-facing error.

The new tests only verify the mapping table, not real driver behavior — manual testing with a wrong URI scheme and an unroutable host would confirm whether these branches are reachable.

server/src/routes/connect.ts
```diff
@@ -66,22 +76,32 @@ type ConnectBody = {
...
+      host,
+      port: effectivePort,
+      user,
+      password: password ?? '',
+      database,
+      ssl,
+      type: dbType,
+      ...(connectionString ? { connectionString } : {}),
+    },
   };
```
eugene-mobile, line 110: **Behavior when both `connectionString` and `user`/`password` are provided is silent.** `MongoDBDriver.buildMongoUri` returns the connectionString verbatim, so per-field credentials are ignored without any indication. Two reasonable options:

1. Reject with 400 (`'Provide either connectionString or user/password, not both.'`) — strictest, easiest to reason about.
2. Document the precedence in the JSDoc on `ConnectionConfig.connectionString` (it already says "instead of building one from host/port/user/password" — fine), AND drop `user`/`password` from the config when `connectionString` is set so they don't end up persisted in saved connections / logged elsewhere.

Related: `postConnect` returns `connectionName: ${user}@${host}:${port}` in the success response. When `connectionString` overrides host/port (e.g. a `mongodb+srv://` URI to a different cluster), this label is misleading — the user thinks they connected to `h:27017` but they're really on `cluster0.example.net`. Consider parsing the URI for display, or just showing `${user}@<connectionString>` with the password stripped.

server/src/routes/connect.test.ts
```diff
@@ -55,6 +55,31 @@ describe('postConnect — db type plumbing', () => {
...
+      .send({ host: 'h', user: 'u', password: 'p', type: 'mongodb', connectionString: uri });
+    expect(res.status).toBe(200);
+    expect(connect).toHaveBeenCalledWith(expect.objectContaining({ type: 'mongodb', connectionString: uri }));
+  });
+
+  it("does not include connectionString when omitted", async () => {
+    await request(makeApp())
+      .post('/api/connect')
+      .send({ host: 'h', user: 'u', password: 'p', type: 'mongodb' });
+    const cfg = (connect as unknown as { mock: { calls: unknown[][] } }).mock.calls[0]?.[0] as Record<string, unknown>;
```
eugene-mobile, line 80: **Test gap: `connectionString` passthrough is only verified at the route boundary.** This asserts the route forwards `connectionString` to `connect()`, but `connect()` is mocked. The actual contract — `MongoDBDriver.buildMongoUri` short-circuits on `config.connectionString` and ignores `host`/`port`/`user`/`password` — has no direct unit test in this PR or in `mongodb.ts`'s test file (if one exists).

A small driver-level test would close the loop:
```ts
expect(buildMongoUri({ connectionString: 'mongodb+srv://...' , host: 'ignored', ... })).toBe('mongodb+srv://...');
```
Non-blocking for this PR if the follow-up PRs add it, but worth a TODO.

server/src/routes/connect.ts
```diff
@@ -66,22 +76,32 @@ type ConnectBody = {
...
   const dbType: DbType = (type as DbType | undefined) ?? 'mysql';
   const effectivePort = Number(port) || defaultPort(dbType);
   return {
-    config: { host, port: effectivePort, user, password: password ?? '', database, ssl, type: dbType },
+    config: {
+      host,
+      port: effectivePort,
+      user,
+      password: password ?? '',
+      database,
```
eugene-mobile: Minor: `...(connectionString ? { connectionString } : {})` treats `''` and `null` and `undefined` identically (all excluded), which is the right behavior, but the spread-conditional pattern is slightly awkward. Equivalent and clearer:
```ts
const config: ConnectionConfig = { host, port: effectivePort, user, password: password ?? '', database, ssl, type: dbType };
if (connectionString) config.connectionString = connectionString;
return { config };
```
Non-blocking style nit.

eugene-mobile (commented):
Solid, well-scoped wiring PR — route allowlist, default port, factory, and error labels all line up. A few notes on the `friendlyConnectError` precedence and on whether the new MongoDB codeName mappings actually fire in practice; nothing blocking.

server/src/routes/connect.ts
eugene-mobile: Done in da86116 — flipped to `e?.codeName ?? e?.code` and added a regression test asserting the lookup hits when both `code` (numeric) and `codeName` are present.

server/src/routes/connect.ts
eugene-mobile: Done in da86116 — dropped `HostUnreachable` and `InvalidURI` from `byCode`; `AuthenticationFailed` stays. Added an inline comment explaining why those reply-codes don't surface in practice. The corresponding tests in `connect.test.ts` were removed.

server/src/routes/connect.ts
eugene-mobile: Done in da86116. `parseConnectBody` now returns 400 (`'Provide either connectionString or user/password, not both.'`) when either credential field is supplied alongside `connectionString`. Also fixed the misleading `connectionName`: when `connectionString` is set we parse the URI for the label (host + optional port + decoded username) and never include the password. SRV URIs omit the port since the real one is resolved via DNS.

server/src/routes/connect.test.ts
eugene-mobile: Done in da86116 — added a driver-level test in `mongodb.test.ts` asserting that when `connectionString` is set, the `MongoClient` ctor receives the URI verbatim and `host`/`port`/`user`/`password` from the config are ignored.

server/src/routes/connect.ts
eugene-mobile: Deferring this one as you noted it's non-blocking style. The conditional-spread is consistent with how we exclude other optional fields and the behavior (treat `''` / `null` / `undefined` identically) is the right one. Happy to fold the rewrite into a follow-up sweep if it bothers you in review elsewhere.

eugene-mobile (commented):
Re-review of da86116 — all five comments resolved correctly. Two small follow-ups noted inline (multi-host URI fallback, friendlyConnectError host context when connectionString is set); both non-blocking.

server/src/routes/connect.ts
```diff
@@ -66,23 +78,65 @@ type ConnectBody = {
...
+ * if the URI can't be parsed (e.g. `mongodb+srv://` in older Node URL impls).
+ */
+function connectionLabel(config: ConnectionConfig): string {
+  if (config.connectionString) {
+    try {
+      // URL doesn't natively understand mongodb+srv; rewrite the scheme so the
+      // parser populates host/username. Strip the password from display.
+      const isSrv = /^mongodb\+srv:\/\//.test(config.connectionString);
+      const normalized = config.connectionString.replace(/^mongodb\+srv:\/\//, 'mongodb://');
+      const url = new URL(normalized);
```
eugene-mobile, line 127: Non-blocking: legitimate replica-set URIs of the form `mongodb://host1:27017,host2:27018/db` make `new URL(...)` throw (the comma in the host list isn't valid per WHATWG), so the label falls back to the literal `<connectionString>` placeholder. Graceful enough for now, but if MongoDB Atlas non-SRV URIs become common in this app, consider extracting the first host with a regex before handing off to `URL`. No action required for this PR.

server/src/routes/connect.ts
```diff
@@ -95,7 +149,7 @@ export const postConnect: RequestHandler = async (req, res) => {
 
   try {
     await connect(config);
-    res.json({ ok: true, connectionName: `${config.user}@${config.host}:${config.port}` });
+    res.json({ ok: true, connectionName: connectionLabel(config) });
   } catch (err) {
     res.status(400).json({ error: friendlyConnectError(err, config.host, config.port, config.type) });
```
eugene-mobile, line 154: Non-blocking: when `connectionString` is supplied without `host`, `config.host` is `''` and `config.port` defaults to 27017, so a network error renders as `Couldn't reach the database at :27017.` — slightly confusing. Could either parse host out of the URI for the error context too, or special-case the message when `connectionString` is set. Worth a follow-up but not blocking.

Issue #113, opened 2026-04-29
ConnectionManager: MongoDB option + connection string toggle


eugene-mobile:
Part of the MongoDB Support project (https://github.com/orgs/appspace/projects/2).

## Task
Update `src/components/ConnectionManager.tsx` to add MongoDB to the DB type selector and offer a connection-string input.

## Acceptance Criteria
- [ ] Database-type selector includes a 'MongoDB' option alongside MySQL / PostgreSQL
- [ ] Default port auto-fills to 27017 when MongoDB is selected (preserving the user-typed-port heuristic from #99)
- [ ] When MongoDB is selected: a toggle switches between **individual fields** (host / port / user / password / database) and **connection string** input
- [ ] Connection-string input is forwarded to the backend as `connectionString`
- [ ] Modal subtitle updates to "Connect to a MongoDB server"

PR #120, opened 2026-04-29
feat: savedConnections - add mongodb type and connectionString (#112)

### LINKED ISSUE #112: savedConnections: add mongodb type and connectionString

## Task
Update `src/savedConnections.ts` to persist MongoDB connections.

## Acceptance Criteria
- [ ] `'mongodb'` added to the `type` union on `SavedConnection`
- [ ] Optional `connectionString?: string` field added
- [ ] Existing saved entries (no `type`) continue to default to `'mysql'`
- [ ] localStorage round-trip remains stable (no migration needed)

eugene-mobile:
## Summary

Extends `SavedConnection` (frontend localStorage helper) so MongoDB entries can be persisted alongside existing MySQL/PostgreSQL ones. Backend wiring for `'mongodb'` already landed in PR #119.

Closes #112

## Changes

- `src/savedConnections.ts` — add `'mongodb'` to the optional `type` union, add optional `connectionString?: string` (with a JSDoc invariant noting it is mongodb-only).
- `src/components/ConnectionManager.tsx` — extract a `formType()` helper (exhaustive switch over `'mysql' | 'postgres' | 'mongodb' | undefined`, parameter renamed from `t` so it doesn't shadow the `t: Theme` prop) at the two existing `entry.type ?? 'mysql'` sites so the wider union narrows cleanly to the form's `'mysql' | 'postgres'`. MongoDB-typed saved entries fall back to the mysql form for now; the full mongodb form UI lands in #113.
- `src/components/ConnectionManager.tsx` — carry mongodb `connectionString` opaquely from the saved entry through the form via a name-keyed `useRef<Map>`, seeded from `applySaved` and from the initial saved entry, then augment the submit payload so re-save preserves the field through the UI flow. `App.handleConnect` forwards `form.connectionString` into both the `/api/connect` body and the `saveConnection()` call.
- `src/components/ConnectionManager.tsx` — autocomplete row only renders the `user@host:port` line when at least one of those is truthy. For mongodb-only entries (where only `connectionString` is set) it falls back to the URL host extracted from the connection string, avoiding a bare `@:` rendering.

## Acceptance criteria

- [x] `'mongodb'` added to the `type` union on `SavedConnection`
- [x] Optional `connectionString?: string` field added with mongodb-only JSDoc invariant
- [x] Existing saved entries (no `type`) continue to default to `'mysql'`
- [x] localStorage round-trip remains stable (no migration needed)
- [x] **UI re-save round-trip** preserves `connectionString` (storage round-trip already worked; the second commit also makes the in-memory form carry it through onConnect → saveConnection)

## Test plan

- [x] `npm run build` (root: tsc + Vite) — clean
- [x] `cd server && npm run build` — clean (no server changes, sanity only)
- [x] Logic smoke: legacy entry without `type` resolves to `mysql`; postgres entry preserved; mongodb entry round-trips with `connectionString` intact through both storage and the UI form.
- [x] Browser smoke (mongodb-only entry injected via DevTools): autocomplete row renders the URL host instead of `@:`; clicking Connect emits a `/api/connect` body containing `connectionString`; localStorage entry retains `connectionString` after the failed connect attempt.
- [ ] Manual: open the app, save a connection (mysql or postgres), refresh, confirm it loads. Older entries (no `type`) still load as mysql.

### DIFF
--- a/src/components/ConnectionManager.tsx
+++ b/src/components/ConnectionManager.tsx
@@ -30,8 +30,9 @@ interface ConnectionManagerProps {
 
 export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t }: ConnectionManagerProps) {
   const [saved, setSaved] = useState<SavedConnection[]>(() => listSavedConnections());
+  const formType = (t: SavedConnection['type']): 'mysql' | 'postgres' => t === 'postgres' ? 'postgres' : 'mysql';
   const initialForm: ConnectionForm = saved[0]
-    ? { ...saved[0], type: saved[0].type ?? 'mysql', password: '', sslVerify: saved[0].sslVerify ?? false, savePassword: saved[0].savePassword ?? false }
+    ? { ...saved[0], type: formType(saved[0].type), password: '', sslVerify: saved[0].sslVerify ?? false, savePassword: saved[0].savePassword ?? false }
     : {
         name: 'Local MySQL',
         type: 'mysql',
@@ -110,7 +111,7 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
   };
 
   const applySaved = (entry: SavedConnection) => {
-    setForm({ ...entry, type: entry.type ?? 'mysql', password: '', sslVerify: entry.sslVerify ?? false, savePassword: entry.savePassword ?? false });
+    setForm({ ...entry, type: formType(entry.type), password: '', sslVerify: entry.sslVerify ?? false, savePassword: entry.savePassword ?? false });
     setAppliedSaved(entry.name);
     setSuggestOpen(false);
     if (entry.savePassword && electronAPI) {
--- a/src/components/ConnectionManager.tsx
+++ b/src/components/ConnectionManager.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import type { CSSProperties } from 'react';
 import type { Theme } from '../theme';
 import { api } from '../api';
@@ -18,6 +18,10 @@ export interface ConnectionForm {
   // Only meaningful in Electron — when true, the password is persisted via
   // safeStorage and reloaded the next time this connection is opened.
   savePassword: boolean;
+  // Opaque pass-through for mongodb-only entries — carried from a saved entry
+  // so re-save round-trips it. Not editable in this form; #113 will replace
+  // this with a real mongodb form and remove this field.
+  connectionString?: string;
 }
 
 interface ConnectionManagerProps {
@@ -30,9 +34,20 @@ interface ConnectionManagerProps {
 
 export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t }: ConnectionManagerProps) {
   const [saved, setSaved] = useState<SavedConnection[]>(() => listSavedConnections());
-  const formType = (t: SavedConnection['type']): 'mysql' | 'postgres' => t === 'postgres' ? 'postgres' : 'mysql';
+  // Carries opaque mongodb connectionStrings from saved entries through the form
+  // so re-save preserves them. Keyed by connection name.
+  const connectionStringsRef = useRef<Map<string, string>>(new Map());
+  const formType = (saved: SavedConnection['type']): 'mysql' | 'postgres' => {
+    switch (saved) {
+      case 'postgres': return 'postgres';
+      case 'mysql':
+      case 'mongodb':
+      case undefined:
+        return 'mysql';
+    }
+  };
   const initialForm: ConnectionForm = saved[0]
-    ? { ...saved[0], type: formType(saved[0].type), password: '', sslVerify: saved[0].sslVerify ?? false, savePassword: saved[0].savePassword ?? false }
+    ? { ...saved[0], type: formType(saved[0].type), password: '', sslVerify: saved[0].sslVerify ?? false, savePassword: saved[0].savePassword ?? false, connectionString: saved[0].connectionString }
     : {
         name: 'Local MySQL',
         type: 'mysql',
@@ -42,6 +57,12 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
         password: '', database: '', ssl: false, sslVerify: true, savePassword: false,
       };
   const [form, setForm] = useState<ConnectionForm>(initialForm);
+  // Seed the carry-forward map from any saved entry that already has a connectionString.
+  if (connectionStringsRef.current.size === 0) {
+    for (const entry of saved) {
+      if (entry.connectionString) connectionStringsRef.current.set(entry.name, entry.connectionString);
+    }
+  }
   // Track the saved entry currently reflected in the form, so we only auto-populate once per match.
   const [appliedSaved, setAppliedSaved] = useState<string>(saved[0]?.name ?? '');
   const [suggestOpen, setSuggestOpen] = useState(false);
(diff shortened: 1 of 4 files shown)


src/components/ConnectionManager.tsx
```diff
@@ -30,8 +30,9 @@ interface ConnectionManagerProps {
 
 export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t }: ConnectionManagerProps) {
   const [saved, setSaved] = useState<SavedConnection[]>(() => listSavedConnections());
+  const formType = (t: SavedConnection['type']): 'mysql' | 'postgres' => t === 'postgres' ? 'postgres' : 'mysql';
```
eugene-mobile: **Blocking-ish: parameter `t` shadows the `t: Theme` prop** declared on line 31. It happens to work because `formType` doesn't reference the theme, but this file uses `t` ~50 times for theming and the shadow is easy to extend by accident. Rename the parameter:

```ts
const formType = (saved: SavedConnection['type']): 'mysql' | 'postgres' =>
  saved === 'postgres' ? 'postgres' : 'mysql';
```

While here: the helper's structure (`=== 'postgres' ? 'postgres' : 'mysql'`) silently maps `'mongodb'` and any future variant to `'mysql'` with no compile error. When #113 widens `ConnectionForm['type']` to include `'mongodb'`, this fallback will keep coercing mongodb saved entries into the mysql form even after the real form lands, because the return type still type-checks. Consider an exhaustive switch so the next union extension is a build error rather than a silent demotion:

```ts
const formType = (saved: SavedConnection['type']): 'mysql' | 'postgres' => {
  switch (saved) {
    case 'postgres': return 'postgres';
    case 'mongodb': return 'mysql'; // TODO(#113): route to mongodb form
    case 'mysql': case undefined: return 'mysql';
  }
};
```

src/components/ConnectionManager.tsx
```diff
@@ -110,7 +111,7 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
   };
 
   const applySaved = (entry: SavedConnection) => {
-    setForm({ ...entry, type: entry.type ?? 'mysql', password: '', sslVerify: entry.sslVerify ?? false, savePassword: entry.savePassword ?? false });
+    setForm({ ...entry, type: formType(entry.type), password: '', sslVerify: entry.sslVerify ?? false, savePassword: entry.savePassword ?? false });
```
eugene-mobile: **Blocking: `connectionString` is dropped on apply.** `setForm({ ...entry, ... })` spreads `entry.connectionString` into state, but `ConnectionForm` doesn't declare the field, so it's not actually preserved on the typed form value, and even if it were, no input writes back to it. Combined with the App.tsx side — `App.handleConnect` builds the saved-entry payload from `form.{type,host,port,user,database,ssl,sslVerify,savePassword}` only and never reads `connectionString` — clicking a mongodb-typed saved entry and pressing Connect will write a new saved entry **without** `connectionString`. The field is lost on first re-save.

The PR description claims `connectionString` round-trips through localStorage intact; that's only true if the user never re-saves through the form, which is the common path. For this PR's interim scope (mongodb falls back to mysql form), the safest fix is to **preserve `connectionString` opaquely** through the form lifecycle — e.g. carry it on a hidden field on `ConnectionForm` and have `App.handleConnect` re-emit it on save, or stash it in a ref keyed by `entry.name`. Otherwise #113 inherits a silently-corrupted persisted shape.

src/savedConnections.ts
```diff
@@ -9,6 +9,8 @@ export interface SavedConnection {
   sslVerify?: boolean;
   // Whether the password is stored in the OS keychain (Electron only).
   savePassword?: boolean;
+  // Optional MongoDB connection string (mongodb:// or mongodb+srv://).
+  connectionString?: string;
```
eugene-mobile, line 16: Non-blocking: `connectionString` is now part of the shape for **every** saved entry, including mysql/postgres. `saveConnection` in `App.tsx` doesn't set it for those types (it's `undefined`, which `JSON.stringify` drops), but if any future writer ever populates this on a non-mongodb entry it will be silently persisted. Consider documenting the invariant in the JSDoc — e.g. `// Set only when type === 'mongodb'.` — and ideally model the shape as a discriminated union once #113 lands so `connectionString` is structurally tied to `type: 'mongodb'`.

eugene-mobile (commented):
Small change, mostly correct and well-scoped. Two real issues worth addressing before merge: (1) `connectionString` is not actually round-tripped — `applySaved` doesn't load it into the form and `App.tsx`'s `saveConnection` call doesn't write it, so the field silently drops the moment a mongodb entry is re-saved through the UI; (2) the `formType` helper shadows the `t: Theme` prop and will become a silent foot-gun once #113 widens the form union. Also: the autocomplete row at `ConnectionManager.tsx:271` (outside this diff) unconditionally renders `{c.user}@{c.host}:{c.port}` — the PR description claims "host/port omitted when empty" was verified during smoke testing, but the code has no such conditional. The mongodb entries used in the smoke test must have had host/port populated; a connection-string-only mongodb entry will display `@:`. Not blocking for this PR, but the description shouldn't claim behavior the code doesn't implement.

src/components/ConnectionManager.tsx
eugene-mobile: Fixed in 809fee6. Renamed the parameter from `t` to `saved` and replaced the ternary with an exhaustive switch over `'mysql' | 'postgres' | 'mongodb' | undefined`.

src/components/ConnectionManager.tsx
eugene-mobile: Fixed in 809fee6. Added `connectionString?: string` to `ConnectionForm` as an opaque pass-through, populated a name-keyed `useRef<Map>` from `applySaved` (and from the initial saved entry), and on submit augmented the `onConnect` payload with the carried value. `App.handleConnect` forwards it into both the `/api/connect` body and the `saveConnection()` call so it round-trips through the UI re-save flow now, not just storage. #113 will replace this with a real mongodb form.

src/savedConnections.ts
eugene-mobile: Addressed in 809fee6 with a JSDoc comment documenting the mongodb-only invariant: non-mongodb writers MUST omit the field. Agreed that #113 should tighten this further with a discriminated union once the mongodb form lands.

src/components/ConnectionManager.tsx
```diff
@@ -181,7 +204,12 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
         </div>
 
         <form
-          onSubmit={(e) => { e.preventDefault(); if (!isConnecting) onConnect(form); }}
+          onSubmit={(e) => {
+            e.preventDefault();
+            if (isConnecting) return;
+            const carried = connectionStringsRef.current.get(form.name.trim());
+            onConnect(carried ? { ...form, connectionString: carried } : form);
```
eugene-mobile, line 211: Non-blocking: the submit handler reads `connectionString` from `connectionStringsRef` and ignores `form.connectionString`, even though `applySaved` and `initialForm` both write the field into the form. That gives you two carry mechanisms doing overlapping work. The ref is the authoritative one (it's what reaches `onConnect`); the form field is effectively dead state on this path.

Given #113 will replace ConnectionForm with a real mongodb form, the cleanest follow-up is either: (a) drop `connectionString` from `ConnectionForm` and rely solely on the ref (matches the original plan), or (b) drop the ref and read `form.connectionString` at submit. Picking (a) keeps the form shape minimal until #113 lands. Not worth blocking on, but worth a TODO comment so the redundancy doesn't outlive the migration.

src/components/ConnectionManager.tsx
```diff
@@ -41,6 +57,12 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
         password: '', database: '', ssl: false, sslVerify: true, savePassword: false,
       };
   const [form, setForm] = useState<ConnectionForm>(initialForm);
+  // Seed the carry-forward map from any saved entry that already has a connectionString.
+  if (connectionStringsRef.current.size === 0) {
+    for (const entry of saved) {
+      if (entry.connectionString) connectionStringsRef.current.set(entry.name, entry.connectionString);
+    }
+  }
```
eugene-mobile, line 65: Minor: this seeding block runs on every render (guarded by `size === 0`, so it only populates once, but the loop check still executes each render). Two small concerns: (1) it's a side-effect during render — would be more idiomatic as a `useState` initializer or a `useRef(() => ...)` lazy init; (2) if the saved-connections list is later mutated such that an entry no longer has a `connectionString`, the stale value lingers in the ref because the guard never re-fires. Today nothing deletes connectionStrings, so this is theoretical, but worth folding the seeding into the `saved` useState initializer for clarity.

eugene-mobile (commented):
Follow-up commit 809fee6 addresses all four review items correctly. Approving with two non-blocking notes about the dual carry mechanism and a render-time seeding loop. Both can be cleaned up in #113 when ConnectionForm is replaced with a real mongodb form.

PR #121, opened 2026-04-30
feat: route MQL bodies to driver in mql mode (#109)

### LINKED ISSUE #109: query.ts: route MQL bodies to the driver in mql mode

eugene-mobile:
## Summary

- Branch on `driver.queryMode` in `postQuery`: SQL drivers continue to receive the raw `sql` string from `req.body.sql`; MQL drivers receive a JSON-encoded request object serialized from `req.body.mql`.
- Return **400** with a clear message on mode mismatch in both directions: a SQL string sent to a MongoDB connection, or an MQL object sent to a SQL connection.
- Tests cover SQL happy path (existing), MQL happy path (find + insertOne), validation, driver errors, and both mode-mismatch directions.

## Acceptance Criteria

- [x] When the active driver's `queryMode === 'mql'`, the request body is parsed as a MQL object and forwarded to `driver.query()` (JSON-stringified, matching the driver's parser contract).
- [x] When `queryMode === 'sql'`, current SQL-string behaviour is unchanged.
- [x] 400 with a clear error message when a SQL string is sent to a MQL-mode driver (and vice versa).
- [x] Unit tests cover both modes.

## Notes on driver wiring

The route uses `getDriver()` from `server/src/db.ts` — the existing singleton accessor that all other routes (`schema.ts`, `updateCell.ts`, `insertRow.ts`, `deleteRow.ts`, `tableDdl.ts`, `dropTable.ts`, `mcp.ts`) already use. There is no per-session driver concept yet; `connect.ts` (PR #119) replaces the singleton on every `/api/connect`. So branching on `getDriver().queryMode` is the right hook today and will keep working once a real mongodb connection is opened. No `getCurrentDriver()` equivalent exists.

## Test plan

- [x] `cd server && npm test` — 83/83 passing (5 new MQL/mismatch cases + existing SQL cases updated to include `queryMode: 'sql'` on the mock)
- [ ] Manual verification deferred — requires a live MongoDB connection wired through the UI; covered by integration work in #110/#111.

Closes #109

### DIFF
--- a/server/src/routes/query.test.ts
+++ b/server/src/routes/query.test.ts
@@ -209,6 +209,35 @@ describe('postQuery – mql mode', () => {
     expect(res.body.error).toMatch(/mql/i);
   });
 
+  it.each([
+    ['array', [1, 2]],
+    ['null', null],
+    ['number', 42],
+  ])('returns 400 when mql is %s (non-object payload)', async (_label, payload) => {
+    const mockDriver = { queryMode: 'mql' as const, query: vi.fn() };
+    vi.mocked(getDriver).mockReturnValue(mockDriver as any);
+
+    const res = await request(makeApp())
+      .post('/api/query')
+      .send({ mql: payload });
+
+    expect(res.status).toBe(400);
+    expect(res.body.error).toMatch(/mql/i);
+    expect(mockDriver.query).not.toHaveBeenCalled();
+  });
+
+  it('returns 400 when mql is a string (treated as SQL/MQL mismatch)', async () => {
+    const mockDriver = { queryMode: 'mql' as const, query: vi.fn() };
+    vi.mocked(getDriver).mockReturnValue(mockDriver as any);
+
+    const res = await request(makeApp())
+      .post('/api/query')
+      .send({ mql: 'oops' });
+
+    expect(res.status).toBe(400);
+    expect(mockDriver.query).not.toHaveBeenCalled();
+  });
+
   it('returns 400 when driver.query() throws in mql mode', async () => {
     const mockDriver = {
       queryMode: 'mql' as const,
@@ -253,6 +282,32 @@ describe('postQuery – mode mismatch', () => {
     expect(res.status).toBe(400);
     expect(res.body.error).toMatch(/SQL mode/i);
     expect(res.body.error).toMatch(/MQL request/i);
+    expect(mockDriver.query).not.toHaveBeenCalled();
+  });
+
+  it('returns 400 when both sql and mql are sent in sql mode (mismatch wins over valid sql)', async () => {
+    const mockDriver = { queryMode: 'sql' as const, query: vi.fn() };
+    vi.mocked(getDriver).mockReturnValue(mockDriver as any);
+
+    const res = await request(makeApp())
+      .post('/api/query')
+      .send({ sql: 'SELECT 1', mql: { collection: 'x', operation: 'find' } });
+
+    expect(res.status).toBe(400);
+    expect(res.body.error).toMatch(/SQL mode/i);
+    expect(mockDriver.query).not.toHaveBeenCalled();
+  });
+
+  it('returns 400 when both sql and mql are sent in mql mode (mismatch wins over valid mql)', async () => {
+    const mockDriver = { queryMode: 'mql' as const, query: vi.fn() };
+    vi.mocked(getDriver).mockReturnValue(mockDriver as any);
+
+    const res = await request(makeApp())
+      .post('/api/query')
+      .send({ sql: 'SELECT 1', mql: { collection: 'x', operation: 'find' } });
+
+    expect(res.status).toBe(400);
+    expect(res.body.error).toMatch(/MQL mode/i);
     expect(mockDriver.query).not.toHaveBeenCalled();
   });
 });
--- a/server/src/routes/query.test.ts
+++ b/server/src/routes/query.test.ts
@@ -20,11 +20,12 @@ function makeApp() {
   return app;
 }
 
-describe('postQuery – driver delegation', () => {
+describe('postQuery – driver delegation (sql mode)', () => {
   beforeEach(() => vi.clearAllMocks());
 
   it('passes schema to driver.query() and returns rows + columnMeta', async () => {
     const mockDriver = {
+      queryMode: 'sql' as const,
       query: vi.fn().mockResolvedValue({
         rows: [{ id: 1, name: 'Alice' }],
         columnMeta: [makeMeta('id', { pk: true }), makeMeta('name')],
@@ -46,6 +47,7 @@ describe('postQuery – driver delegation', () => {
 
   it('passes undefined schema when no schema is provided', async () => {
     const mockDriver = {
+      queryMode: 'sql' as const,
       query: vi.fn().mockResolvedValue({
         rows: [{ one: 1 }],
         columnMeta: [makeMeta('one')],
@@ -63,6 +65,7 @@ describe('postQuery – driver delegation', () => {
 
   it('returns affectedRows and insertId for DML (empty columnMeta)', async () => {
     const mockDriver = {
+      queryMode: 'sql' as const,
       query: vi.fn().mockResolvedValue({
         rows: [],
         columnMeta: [],
@@ -84,6 +87,7 @@ describe('postQuery – driver delegation', () => {
 
   it('returns 400 when driver.query() throws', async () => {
     const mockDriver = {
+      queryMode: 'sql' as const,
       query: vi.fn().mockRejectedValue(new Error("Table 'mydb.ghost' doesn't exist")),
     };
     vi.mocked(getDriver).mockReturnValue(mockDriver as any);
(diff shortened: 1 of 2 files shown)


eugene-mobile (commented):
Mode-dispatch logic is correct (driver.queryMode-based, not body-shape-based) and the test coverage is solid. A few non-blocking observations on input validation tightness and error-status semantics that you may want to follow up on, plus two small test gaps.

server/src/routes/query.ts
```diff
@@ -2,16 +2,45 @@ import type { RequestHandler } from 'express';
...
+    // mql mode
+    if (typeof sql === 'string') {
+      res.status(400).json({
+        error: 'This connection is in MQL mode; expected a "mql" object body, not a SQL string.',
+      });
+      return;
+    }
+    if (!mql || typeof mql !== 'object') {
+      res.status(400).json({ error: 'mql request object is required.' });
+      return;
```
eugene-mobile, line 39: **Non-blocking — looser than necessary.** `typeof mql === 'object'` admits arrays and other non-plain objects (since `typeof [] === 'object'` and `![]` is `false`). An array sent as `mql` will be `JSON.stringify`'d and forwarded; `MongoDBDriver.parseRequest` will then reject it because `parsed.collection` is missing, surfacing the message `MQL request requires a "collection" field.` rather than something like `mql must be an object`. Consider tightening the check to `Array.isArray(mql) || typeof mql !== 'object'` here so the route returns a more accurate 400 for arrays/non-objects without invoking the driver. Same trade-off you already make for `sql` (you reject blank strings up-front instead of letting the driver complain).

server/src/routes/query.ts
```diff
@@ -2,16 +2,45 @@ import type { RequestHandler } from 'express';
...
+    mql?: unknown;
+    schema?: string;
+  };
 
-  if (!sql?.trim()) {
-    res.status(400).json({ error: 'sql is required.' });
-    return;
+  let queryText: string;
+
+  if (driver.queryMode === 'sql') {
```
eugene-mobile, line 17: **Non-blocking — undocumented behaviour.** In SQL mode, `mql !== undefined` means a body containing **both** `sql` and `mql` (or even `mql: null`) is rejected with the mode-mismatch error instead of processing the valid `sql`. Symmetric in MQL mode (`typeof sql === 'string'` rejects even when `mql` is also present). This is defensible (it catches confused clients) but the current PR description and code comments don't say so — worth a one-line comment above the check, e.g. `// Both fields present is treated as a mismatch — the client is clearly confused about the connection mode.` so a future reader doesn't "fix" it.

server/src/routes/query.ts
```diff
@@ -2,16 +2,45 @@ import type { RequestHandler } from 'express';
...
+      return;
+    }
+    queryText = JSON.stringify(mql);
   }
 
   try {
     const start = Date.now();
-    const result = await getDriver().query(sql, [], schema);
+    const result = await driver.query(queryText, [], schema);
     const executionTime = Date.now() - start;
```
eugene-mobile, line 47: **Non-blocking, follow-up.** All `driver.query()` rejections map to **400** here — that's preserved from the pre-existing SQL behaviour, so this PR isn't introducing the issue. But MQL mode widens the surface: `MongoServerSelectionError`, `MongoNetworkError`, and connect-time failures inside `ensureConnected()` are infrastructure problems, not client mistakes, and they will now be reported as 400. Worth a follow-up issue to classify driver errors into 4xx (parse / unsupported operation / validation) vs 5xx (connection / timeout / internal) — probably easier once both drivers expose structured error codes.

server/src/routes/query.ts
```diff
@@ -2,16 +2,45 @@ import type { RequestHandler } from 'express';
 import { getDriver } from '../db.js';
 
 export const postQuery: RequestHandler = async (req, res) => {
-  const { sql, schema } = req.body as { sql: string; schema?: string };
+  const driver = getDriver();
+  const { sql, mql, schema } = req.body as {
+    sql?: unknown;
```
eugene-mobile, line 7: **Minor — input typing asymmetry.** `sql` and `mql` are typed as `unknown` and validated at runtime, but `schema` is typed as `string` and trusted. A client sending `schema: 123` or `schema: { $gt: '' }` will pass it straight to `driver.query()`, which in turn passes it to `MongoDBDriver.db(schema)` where `this.client.db(name)` will throw a non-obvious error. Pre-existing for the SQL path (mysql2 just stringifies), but MQL mode is a good moment to add `typeof schema !== 'string' && schema !== undefined` validation — purely a robustness/error-message improvement, no security impact since the driver layer rejects it.

server/src/routes/query.test.ts
```diff
@@ -136,3 +150,109 @@ describe('postQuery – row serialization passthrough', () => {
...
+
+    const res = await request(makeApp())
+      .post('/api/query')
+      .send({ mql: { collection: 'users', operation: 'find' } });
+
+    expect(res.status).toBe(400);
+    expect(res.body.error).toMatch(/SQL mode/i);
+    expect(res.body.error).toMatch(/MQL request/i);
+    expect(mockDriver.query).not.toHaveBeenCalled();
+  });
```
eugene-mobile, line 286: **Test gap.** No case exercises a body with **both** `sql` and `mql` present. Given the route's current behaviour (in SQL mode, any defined `mql` triggers mismatch even alongside a valid `sql`; symmetric in MQL mode), this is worth pinning down with a test in each mode so a future refactor doesn't silently change the contract. Two small additions:

```ts
it('returns 400 when both sql and mql are sent in sql mode', async () => {
  vi.mocked(getDriver).mockReturnValue({ queryMode: 'sql' as const, query: vi.fn() } as any);
  const res = await request(makeApp())
    .post('/api/query')
    .send({ sql: 'SELECT 1', mql: { collection: 'x', operation: 'find' } });
  expect(res.status).toBe(400);
});
```

and the symmetric MQL-mode case.

server/src/routes/query.test.ts
```diff
@@ -136,3 +150,109 @@ describe('postQuery – row serialization passthrough', () => {
...
+    expect(res.body.affectedRows).toBe(1);
+    expect(res.body.rows).toEqual([]);
+  });
+
+  it('returns 400 when mql is missing in mql mode', async () => {
+    vi.mocked(getDriver).mockReturnValue({ queryMode: 'mql' as const, query: vi.fn() } as any);
+    const res = await request(makeApp()).post('/api/query').send({});
+    expect(res.status).toBe(400);
+    expect(res.body.error).toMatch(/mql/i);
+  });
```
eugene-mobile, line 210: **Test gap.** No case for non-object MQL payloads — e.g. `mql: [1, 2]`, `mql: 'string'`, `mql: 42`, `mql: null`. Today an **array** slips past the route's `typeof mql === 'object'` check and is forwarded to the driver (see comment on `query.ts` line 36); a primitive is correctly rejected. Either tighten the route check and add a test asserting 400 with a `not an object` message, or add a test that pins the current behaviour (driver-rejected with `requires a "collection" field`) so it's deliberate.

server/src/routes/query.ts
eugene-mobile: Tightened in cb5105c — the check is now `mql === null || typeof mql !== 'object' || Array.isArray(mql)`, so arrays and `null` are rejected at the route boundary with the cleaner "mql request object is required" message instead of being forwarded to the driver.

server/src/routes/query.ts
eugene-mobile: Added an inline comment in cb5105c above the mode dispatch documenting that "both sql and mql present" is intentionally treated as a mismatch and the mismatch check therefore runs first.

server/src/routes/query.ts
eugene-mobile: Agreed — deferring. Filed #122 to classify driver errors (`client` / `transient` / `server`) once both drivers expose structured error codes; cb5105c adds a code comment near the 400 mapping referencing the issue so the lossy mapping is documented.

server/src/routes/query.ts
eugene-mobile: Rejecting in this PR — `schema` is typed-and-trusted across every route (`/api/schema`, `/api/ddl`, `/api/insert-row`, etc.), not just `/api/query`. A one-off runtime check here would be inconsistent; the right fix is a shared body-validation layer applied to all routes, which is well outside #109's scope. Happy to file a separate issue if you'd like to track it.

server/src/routes/query.test.ts
eugene-mobile: Added in cb5105c — two new tests in the "mode mismatch" describe block pin the contract: in SQL mode and in MQL mode, sending both `sql` and a valid `mql` returns 400 with the mode-mismatch error and the driver is never called.

server/src/routes/query.test.ts
eugene-mobile: Added in cb5105c — `it.each` covers `mql: [1,2]`, `mql: null`, `mql: 42`, plus a separate case for `mql: 'string'`. After the route tightening (same commit), all four return 400 with the "mql request object is required" message and the driver is never invoked.

eugene-mobile (commented):
Re-review of cb5105c: all six prior comments resolved appropriately.

- **(1) Array/null/non-object MQL** — fixed at L37 with explicit `null` / `Array.isArray` guards; tests pin array, null, number, and string payloads.
- **(2) Mode-mismatch precedence** — comment at L14-16 makes the "both fields => 400" contract explicit; two new tests pin it in each direction.
- **(3) 400-vs-500 driver errors** — deferred to #122 with an inline reference at L67-69. Reasonable: real fix needs structured `DriverError` classes on both drivers, out of scope here.
- **(4) `schema` runtime validation** — rejection accepted. Confirmed `updateCell.ts`, `deleteRow.ts`, `insertRow.ts`, `dropTable.ts` all `as`-cast `schema?: string` without runtime validation. Not specific to #109; right fix is a shared body-validation layer (zod or similar) applied uniformly. No security risk here — every consumer wraps `schema` in backticks after stripping backticks.
- **(5) Both-fields test gap** — two tests added (sql-mode and mql-mode), both asserting `driver.query` is not called.
- **(6) Non-object MQL test gap** — `it.each` covers array/null/number plus a separate string case; all assert `driver.query` is not called.

Tests: 89/89 pass. No new issues spotted. LGTM.

Issue #118, opened 2026-04-29
MongoDB: union-of-keys schema sampling for getSchema/getTable


eugene-mobile:
Follow-up from PR #117 review (comment 3164075737).

## Context

`MongoDBDriver.inferColumns` currently samples a single document via `findOne({})` per collection to infer fields. This is fragile for collections with schema variance:
- Fields present only on later documents are invisible.
- The sample is non-deterministic without a sort.
- Field type can vary across documents; we report only what the first sample shows.

We document this as best-effort today by setting `inferred: true` on every column, but the UX is poor for evolving collections.

## Proposed approach

Sample N documents (e.g. 50) and union the keys, recording for each field:
- the set of observed types
- a notNull flag (true when present in 100% of samples)
- the inferred sample size

Sort to make the sample deterministic (e.g. `{_id: -1}` to bias toward recent writes).

## Acceptance criteria

- [ ] `getSchema` and `getTable` sample N documents (configurable, default 50)
- [ ] Field list is the union of keys across the sample
- [ ] `ColumnInfo.dataType` reflects the dominant type when types vary; conflict surfaced via a new field or via `comment`
- [ ] `nullable` is derived from the sample (false only when the field is present in every sampled doc)
- [ ] Existing tests in `server/src/drivers/mongodb.test.ts` still pass

## Out of scope

- Reading the collection's `\$jsonSchema` validator to drive field metadata (separate issue).

Issue #114, opened 2026-04-29
QueryEditor: switch to JSON/MQL mode


eugene-mobile:
Part of the MongoDB Support project (https://github.com/orgs/appspace/projects/2).

## Task
Update `src/components/QueryEditor.tsx` to switch into MQL mode when the active driver reports `queryMode === 'mql'`.

## Acceptance Criteria
- [ ] Editor switches to JSON syntax highlighting (not SQL) when queryMode is `mql`
- [ ] Helper text above the editor shows MQL examples — e.g. `{ "filter": { "age": { "$gt": 30 } } }`
- [ ] Collection selector dropdown populated from the current schema's collections
- [ ] Submitting sends a parsed MQL object to `/api/query` (not a string)
- [ ] SQL behaviour for MySQL/Postgres is unchanged

PR #123, opened 2026-04-30
test: cover schema route MongoDB (mql) path (#111)

### LINKED ISSUE #111: schema.ts: support inferred MongoDB columns

eugene-mobile:
## Summary

Closes #111.

The schema route (`server/src/routes/schema.ts`) was already driver-agnostic — it delegates entirely to `driver.getSchemas()` / `driver.getSchema()`. The MongoDB foundation work in #117 added both methods to `MongoDBDriver` (collection listing + per-collection field sampling with `inferred: true`), so `/api/schemas` and `/api/schema` already return correctly-shaped responses for MongoDB connections.

This PR adds explicit test coverage so the behaviour is locked in and a future refactor can't silently regress the mql path.

## Acceptance criteria

- [x] For MongoDB, `getSchema` returns collections with fields sampled from one document — covered by the `mql mode` describe block (driver mock returns `inferred: true` columns and route passes them through unchanged).
- [x] Column entries include `inferred: true` so the UI can indicate uncertainty — asserted on every column of every table in the mql happy-path test.
- [x] Heterogeneous collections gracefully return whatever fields the sample document had (no error) — covered by the `events` vs `sparse` test where sibling collections expose different column sets.
- [x] No regression for MySQL/Postgres (`inferred` undefined or false) — covered by the sql-mode test asserting `not.toHaveProperty('inferred')`.

## What changed

- New file: `server/src/routes/schema.test.ts` (10 tests, 270 LOC)
- No production code changes — the route is already driver-agnostic and the MongoDB driver already implements the contract.

## Test plan

- [x] `cd server && npm test` — 99/99 green (10 new in `schema.test.ts`)
- [ ] End-to-end smoke against a real MongoDB will land with the frontend MongoDB connection form (#113 / #114). Today the connection UI doesn't expose `type: 'mongodb'`, so a browser smoke isn't yet possible from the UI; the type-checked driver dispatch + unit tests are sufficient.

## Out of scope

- Frontend rendering of MongoDB collections in the sidebar — handled downstream in #113 / #114.
- DDL-equivalent endpoints (`/api/ddl` with a collection) — that's #110.
- Union-of-keys multi-doc field sampling — tracked in #118.

### DIFF
--- a/server/src/routes/schema.test.ts
+++ b/server/src/routes/schema.test.ts
@@ -1,0 +1,270 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import request from 'supertest';
+import express from 'express';
+
+vi.mock('../db.js', () => ({
+  getDriver: vi.fn(),
+}));
+
+import { getDriver } from '../db.js';
+import { getSchemas, getSchema } from './schema.js';
+
+function makeApp() {
+  const app = express();
+  app.use(express.json());
+  app.get('/api/schemas', getSchemas);
+  app.get('/api/schema', getSchema);
+  return app;
+}
+
+function sqlColumn(name: string, opts: Partial<{ pk: boolean; nullable: boolean; type: string }> = {}) {
+  return {
+    name,
+    type: opts.type ?? 'varchar(255)',
+    dataType: opts.type ?? 'varchar',
+    pk: opts.pk ?? false,
+    nullable: opts.nullable ?? true,
+    default: null,
+    autoIncrement: false,
+    comment: '',
+  };
+}
+
+function mqlColumn(name: string, dataType: string, opts: Partial<{ pk: boolean; nullable: boolean }> = {}) {
+  return {
+    name,
+    type: dataType,
+    dataType,
+    pk: opts.pk ?? false,
+    nullable: opts.nullable ?? true,
+    default: null,
+    autoIncrement: false,
+    comment: '',
+    inferred: true,
+  };
+}
+
+describe('getSchemas — driver delegation', () => {
+  beforeEach(() => vi.clearAllMocks());
+
+  it('returns the list of schemas from a sql-mode driver', async () => {
+    const mockDriver = {
+      queryMode: 'sql' as const,
+      getSchemas: vi.fn().mockResolvedValue(['app', 'analytics']),
+    };
+    vi.mocked(getDriver).mockReturnValue(mockDriver as any);
+
+    const res = await request(makeApp()).get('/api/schemas');
+
+    expect(res.status).toBe(200);
+    expect(res.body).toEqual({ schemas: ['app', 'analytics'] });
+    expect(mockDriver.getSchemas).toHaveBeenCalledTimes(1);
+  });
+
+  it('returns the list of databases from a mql-mode driver', async () => {
+    const mockDriver = {
+      queryMode: 'mql' as const,
+      getSchemas: vi.fn().mockResolvedValue(['blog', 'shop']),
+    };
+    vi.mocked(getDriver).mockReturnValue(mockDriver as any);
+
+    const res = await request(makeApp()).get('/api/schemas');
+
+    expect(res.status).toBe(200);
+    expect(res.body).toEqual({ schemas: ['blog', 'shop'] });
+    expect(mockDriver.getSchemas).toHaveBeenCalledTimes(1);
+  });
+
+  it('returns 500 when the driver throws', async () => {
+    const mockDriver = {
+      queryMode: 'sql' as const,
+      getSchemas: vi.fn().mockRejectedValue(new Error('boom')),
+    };
+    vi.mocked(getDriver).mockReturnValue(mockDriver as any);
+
+    const res = await request(makeApp()).get('/api/schemas');
+
+    expect(res.status).toBe(500);
+    expect(res.body.error).toBe('boom');
+  });
+});
+
+describe('getSchema — sql mode', () => {
+  beforeEach(() => vi.clearAllMocks());
+
+  it('returns tables/views/procedures/triggers without inferred flags', async () => {
+    const mockDriver = {
+      queryMode: 'sql' as const,
+      getSchema: vi.fn().mockResolvedValue({
+        tables: [
+          {
+            name: 'users',
+            rows: 42,
+            comment: '',
+            columns: [sqlColumn('id', { pk: true, nullable: false, type: 'int' }), sqlColumn('email')],
+          },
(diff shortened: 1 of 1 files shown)


server/src/routes/schema.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    expect(mockDriver.getSchemas).toHaveBeenCalledTimes(1);
+  });
+
+  it('returns the list of databases from a mql-mode driver', async () => {
+    const mockDriver = {
+      queryMode: 'mql' as const,
+      getSchemas: vi.fn().mockResolvedValue(['blog', 'shop']),
+    };
+    vi.mocked(getDriver).mockReturnValue(mockDriver as any);
+
```
eugene-mobile, line 70: Setting `queryMode: 'mql'` / `'sql'` on the mock driver is decorative — `schema.ts` never reads `driver.queryMode`. The 'sql vs mql' split in this file is really 'fixture with `inferred:true` vs without'. Worth either dropping `queryMode` from the mocks (less confusion about what's being asserted) or adding a comment that the route is mode-agnostic and these describe blocks just exist to pin response shape per fixture variant.

server/src/routes/schema.test.ts
```diff
@@ -0,0 +1,270 @@
...
+              mqlColumn('_id', 'objectId', { pk: true, nullable: false }),
+              mqlColumn('total', 'double'),
+            ],
+          },
+        ],
+        views: [],
+        procedures: [],
+        triggers: [],
+      }),
+    };
```
eugene-mobile, line 158: This loop asserts `inferred: true` on every column, but the test itself constructs those columns via `mqlColumn()` which hard-codes `inferred: true`. Net assertion is 'route doesn't strip fields' — fine as a contract guard, but it doesn't validate that `MongoDBDriver` actually emits `inferred: true` (that lives in `mongodb.test.ts`). Consider a one-line comment so a future reader doesn't think this is end-to-end coverage of the inference contract.

server/src/routes/schema.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+    const res = await request(makeApp()).get('/api/schema').query({ schema: 'app' });
+
+    expect(res.status).toBe(200);
+    expect(mockDriver.getSchema).toHaveBeenCalledWith('app');
+    expect(res.body.tables).toHaveLength(1);
+    expect(res.body.tables[0].columns[0]).not.toHaveProperty('inferred');
+    expect(res.body.views).toEqual(['active_users']);
+    expect(res.body.procedures).toEqual(['cleanup']);
+    expect(res.body.triggers).toEqual(['audit_users']);
```
eugene-mobile, line 122: Nit: PR description AC says 'no regression for MySQL/Postgres (inferred undefined **or false**)', but `not.toHaveProperty('inferred')` only catches the undefined case. If a future driver started emitting `inferred: false` on every SQL column the test would still pass and the UI would (per #117 contract) treat the column as inferred. Either tighten to `expect(c.inferred).toBeUndefined()` or relax the AC wording.

server/src/routes/schema.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      res.body.tables.find((t: { name: string }) => t.name === name).columns.map((c: { name: string }) => c.name);
+    expect(cols('events')).toEqual(['_id', 'type', 'payload']);
+    expect(cols('sparse')).toEqual(['_id']);
+  });
+
+  it('does not crash when a collection has no sample and only the _id placeholder column is returned', async () => {
+    const mockDriver = {
+      queryMode: 'mql' as const,
+      getSchema: vi.fn().mockResolvedValue({
+        tables: [
```
eugene-mobile, line 220: The `getSchemas` describe has a 'driver throws' case but `getSchema` (mql mode) tests the throw path while sql mode does not, and neither tests the `String(err)` branch for a non-Error rejection (e.g. `Promise.reject('plain string')`). The route has that branch — easy to add a one-liner so the fallback is locked in.

server/src/routes/schema.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  it('returns 400 when schema query param is missing', async () => {
+    const res = await request(makeApp()).get('/api/schema');
+    expect(res.status).toBe(400);
+    expect(res.body.error).toMatch(/schema/i);
+  });
+
+  it('returns 400 when schema is blank whitespace', async () => {
+    const res = await request(makeApp()).get('/api/schema').query({ schema: '   ' });
+    expect(res.status).toBe(400);
```
eugene-mobile, line 268: Minor: the blank-whitespace 400 test asserts status only, not the error message. The missing-param test above asserts `/schema/i` on the body — same handler returns the same 400, so an identical body assertion here would be consistent and cheap.

server/src/routes/schema.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  it('handles heterogeneous collections where sampled docs have different fields', async () => {
+    // Best-effort inference: each collection reflects whatever fields the
+    // single sampled document had, so neighbours can disagree on the column set.
+    const mockDriver = {
+      queryMode: 'mql' as const,
+      getSchema: vi.fn().mockResolvedValue({
+        tables: [
+          {
+            name: 'events',
```
eugene-mobile, line 184: On scope vs issue #111 AC: this PR covers the route-shape ACs but the issue's frontend-facing AC ('sidebar populates with collections for a MongoDB connection') is explicitly deferred to #113/#114 in the PR body. That's the right call given the connection UI doesn't yet expose `type: 'mongodb'`, but worth confirming #111 is closed by this PR rather than by the frontend work — otherwise the link in the commit message is misleading.

eugene-mobile (commented):
Approving — a tests-only PR is the right call here, but the coverage is thinner than the PR description implies. The route is a pure passthrough (it never reads `queryMode`), so most assertions verify that the mocked fixture survives a trip through `res.json` rather than testing real route logic. That is still useful as a regression net against future refactors that might transform the response, but a few of these cases duplicate `mongodb.test.ts` without adding constraint. Inline notes below — none blocking.

PR #124, opened 2026-04-30
feat: ConnectionManager UI for MongoDB (#113)

### LINKED ISSUE #113: ConnectionManager: MongoDB option + connection string toggle

eugene-mobile:
Closes #113

## Summary
- Adds **MongoDB** to the DB type selector alongside MySQL and PostgreSQL; default port auto-fills to **27017** while preserving the user-typed-port heuristic from #99.
- When MongoDB is selected, a **Fields / Connection string** toggle appears. URI mode shows a single textarea; host/port/user/password are hidden and stripped from the request body so the route's `connectionString && (user || password)` 400 check (#119) is never tripped. The "Default database" field stays visible in both modes.
- Modal subtitle reads "Connect to a MongoDB server" when mongodb is selected, "Default database" replaces "Default schema" for the optional db hint.

## #120 follow-ups folded in
- **Dropped the dual-carry**: `connectionString` is now a real form-state field, not a `useRef<Map>` shadow. Saved entries are detected by presence of `connectionString` and load straight into URI mode.
- **No more side-effect-during-render**: the render-time Map-seed loop is gone with the ref.
- `api.connect` / `api.testConnection` now share a single `buildConnectBody()` helper so only known fields are serialized — no `mongoMode` or other UI-only props leak into the request body.

## Acceptance criteria
- [x] Database-type selector includes MongoDB
- [x] Default port auto-fills to 27017 when MongoDB is selected (heuristic preserved)
- [x] Toggle switches between fields and connection-string input modes
- [x] Connection-string input is forwarded as `connectionString`
- [x] Modal subtitle updates to "Connect to a MongoDB server"

## Test plan
- [x] `npm run build` from repo root — clean (Vite + tsc)
- [x] Open New connection → MongoDB; verify port placeholder is `27017`
- [x] Toggle Fields ↔ Connection string; fields hide/show; database field stays
- [x] Save a fields-mode mongodb entry to localStorage → reload → loads back into Fields
- [x] Save a uri-mode mongodb entry → reload → loads back into Connection string with the URI preserved
- [x] Submit fields-mode → POST body contains `type: 'mongodb'`, host/port/user/password populated, no `connectionString`
- [x] Submit uri-mode → POST body contains `type: 'mongodb'`, host/port/user empty, `connectionString` set
- [x] Autocomplete row for a uri-only entry shows `127.0.0.1:27017 · mdb` (no `@:` artifact — #120 fix preserved)
- [x] MySQL and PostgreSQL flows visually unchanged (no mode toggle, original labels)

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -43,6 +43,32 @@ export interface DeleteRowWhere {
   value: string | number | boolean | null;
 }
 
+interface ConnectFormInput {
+  type: 'mysql' | 'postgres' | 'mongodb';
+  host: string;
+  port: string;
+  user: string;
+  password: string;
+  database: string;
+  ssl: boolean;
+  sslVerify: boolean;
+  connectionString?: string;
+}
+
+function buildConnectBody(form: ConnectFormInput) {
+  const sslMode = !form.ssl ? undefined : form.sslVerify ? 'verify-full' : 'require';
+  return {
+    type: form.type,
+    host: form.host,
+    port: Number(form.port),
+    user: form.user,
+    password: form.password,
+    database: form.database,
+    ssl: sslMode,
+    ...(form.connectionString ? { connectionString: form.connectionString } : {}),
+  };
+}
+
 async function request<T>(path: string, init?: RequestInit): Promise<T> {
   const res = await fetch(path, { headers: { 'Content-Type': 'application/json' }, ...init });
   const body = await res.json() as T & { error?: string };
@@ -54,21 +80,17 @@ async function request<T>(path: string, init?: RequestInit): Promise<T> {
 }
 
 export const api = {
-  connect(form: { type: 'mysql' | 'postgres'; host: string; port: string; user: string; password: string; database: string; ssl: boolean; sslVerify: boolean; connectionString?: string }) {
-    const { ssl, sslVerify, ...rest } = form;
-    const sslMode = !ssl ? undefined : sslVerify ? 'verify-full' : 'require';
+  connect(form: ConnectFormInput) {
     return request<{ ok: boolean; connectionName: string }>('/api/connect', {
       method: 'POST',
-      body: JSON.stringify({ ...rest, port: Number(form.port), ssl: sslMode }),
+      body: JSON.stringify(buildConnectBody(form)),
     });
   },
 
-  testConnection(form: { type: 'mysql' | 'postgres'; host: string; port: string; user: string; password: string; database: string; ssl: boolean; sslVerify: boolean }) {
-    const { ssl, sslVerify, ...rest } = form;
-    const sslMode = !ssl ? undefined : sslVerify ? 'verify-full' : 'require';
+  testConnection(form: ConnectFormInput) {
     return request<{ ok: boolean }>('/api/connect/test', {
       method: 'POST',
-      body: JSON.stringify({ ...rest, port: Number(form.port), ssl: sslMode }),
+      body: JSON.stringify(buildConnectBody(form)),
     });
   },
 
--- a/src/api.ts
+++ b/src/api.ts
@@ -45,27 +45,38 @@ export interface DeleteRowWhere {
 
 interface ConnectFormInput {
   type: 'mysql' | 'postgres' | 'mongodb';
-  host: string;
-  port: string;
-  user: string;
-  password: string;
+  host?: string;
+  port?: string;
+  user?: string;
+  password?: string;
   database: string;
   ssl: boolean;
   sslVerify: boolean;
   connectionString?: string;
 }
 
+// Presence-based: when `connectionString` is set we omit host/port/user/password
+// entirely so the wire payload reflects the API invariant that the URI is the
+// sole source of credentials. Mirrors the route's collision check in
+// `server/src/routes/connect.ts`.
 function buildConnectBody(form: ConnectFormInput) {
   const sslMode = !form.ssl ? undefined : form.sslVerify ? 'verify-full' : 'require';
+  if (form.connectionString) {
+    return {
+      type: form.type,
+      database: form.database,
+      ssl: sslMode,
+      connectionString: form.connectionString,
+    };
+  }
   return {
     type: form.type,
-    host: form.host,
-    port: Number(form.port),
-    user: form.user,
-    password: form.password,
+    host: form.host ?? '',
+    port: Number(form.port ?? ''),
+    user: form.user ?? '',
+    password: form.password ?? '',
     database: form.database,
     ssl: sslMode,
-    ...(form.connectionString ? { connectionString: form.connectionString } : {}),
   };
 }
 
--- a/src/components/ConnectionManager.tsx
+++ b/src/components/ConnectionManager.tsx
@@ -47,6 +47,22 @@ const dbLabel = (type: DbType): string => {
   return 'MySQL';
 };
 
+// Mirrors `connectionLabel` in `server/src/routes/connect.ts`: rewrites
+// `mongodb+srv://` so WHATWG URL populates host/username, and prefers
+// `hostname` over `host` so we never leak an explicit port. Returns just the
+// host portion (no `user@`) for the autocomplete subtitle. Falls back to a
+// neutral placeholder rather than the raw URI so a malformed entry — or one
+// we can't parse — never displays an embedded password.
+const hostFromConnectionString = (uri: string): string => {
+  try {
+    const normalized = uri.replace(/^mongodb\+srv:\/\//, 'mongodb://');
+    const url = new URL(normalized);
+    return url.hostname || '<connectionString>';
+  } catch {
+    return '<connectionString>';
+  }
+};
+
 const formFromSaved = (entry: SavedConnection): ConnectionForm => {
   const type: DbType = entry.type ?? 'mysql';
   const mongoMode: 'fields' | 'uri' = type === 'mongodb' && entry.connectionString ? 'uri' : 'fields';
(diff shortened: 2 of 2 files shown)


src/components/ConnectionManager.tsx
```diff
@@ -99,11 +117,22 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
     return () => window.removeEventListener('keydown', key);
   }, [onDismiss, isConnecting]);
 
+  const isMongoUri = form.type === 'mongodb' && form.mongoMode === 'uri';
+
+  const buildSubmitForm = (): ConnectionForm => {
+    if (isMongoUri) {
+      // URI mode owns the credentials; the route 400s if both are sent.
+      return { ...form, host: '', port: '', user: '', password: '' };
```
eugene-mobile: Non-blocking, but worth tightening: the URI-mode submit zeroes host/port/user/password to empty strings and `Number('') === 0`, so `buildConnectBody` ships `{port: 0, host: '', user: '', password: ''}` over the wire. This survives the route's collision check (`connectionString && (user || password)`) only because empty-string is falsy, and `Number(port) || defaultPort(...)` rescues `0`. Both are coincidences. If anyone later changes the route check to presence-based (`'user' in body`) or removes the `0 || default` fallback, this silently breaks.

Cleaner shape: omit the fields entirely instead of zeroing them. Either narrow `ConnectFormInput` in `api.ts` so `host`/`port`/`user`/`password` are optional and have `buildConnectBody` skip them when not set (mirroring the existing `connectionString` spread pattern), or build the request payload directly here rather than round-tripping through `ConnectionForm`.

src/api.ts
```diff
@@ -43,6 +43,32 @@ export interface DeleteRowWhere {
...
+  sslVerify: boolean;
+  connectionString?: string;
+}
+
+function buildConnectBody(form: ConnectFormInput) {
+  const sslMode = !form.ssl ? undefined : form.sslVerify ? 'verify-full' : 'require';
+  return {
+    type: form.type,
+    host: form.host,
+    port: Number(form.port),
```
eugene-mobile: Companion to the URI-mode comment in `ConnectionManager.tsx`: `port: Number(form.port)` produces `port: 0` when `form.port === ''` (URI mode). The backend's `Number(port) || defaultPort(dbType)` covers this today, but sending `0` is misleading — a reader of the wire payload could reasonably read it as 'port zero requested'. Consider:

```ts
const portNum = form.port ? Number(form.port) : undefined;
return {
  type: form.type,
  ...(form.host ? { host: form.host } : {}),
  ...(form.user ? { user: form.user } : {}),
  ...(form.password ? { password: form.password } : {}),
  ...(portNum !== undefined ? { port: portNum } : {}),
  database: form.database,
  ssl: sslMode,
  ...(form.connectionString ? { connectionString: form.connectionString } : {}),
};
```

This makes `buildConnectBody` enforce the API-level invariant that a connectionString is the sole source of truth for credentials, instead of relying on `ConnectionManager.buildSubmitForm` to zero them.

src/components/ConnectionManager.tsx
```diff
@@ -117,23 +146,28 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
     setTestResult(null);
   };
 
-  const defaultPort = (type: 'mysql' | 'postgres') => type === 'postgres' ? '5432' : '3306';
-  const setDbType = (next: 'mysql' | 'postgres') => {
+  const setDbType = (next: DbType) => {
     setForm(p => {
       const prevDefault = defaultPort(p.type);
       const portIsPrevDefault = p.port === '' || p.port === prevDefault;
+      const nextMongoMode = next === 'mongodb' ? (p.mongoMode ?? 'fields') : 'fields';
```
eugene-mobile, line 168: Minor: `nextMongoMode = next === 'mongodb' ? (p.mongoMode ?? 'fields') : 'fields'` resets `mongoMode` when switching *away* from mongodb, but leaves `connectionString` in form state untouched. If a user types a URI in URI mode, switches to MySQL, then switches back to mongodb, mongoMode is now `'fields'` but the stale URI is still in form state (hidden). It can't leak to the wire because `buildSubmitForm`'s fields branch strips `connectionString`, so this is purely cosmetic state. Worth either clearing `connectionString` here too, or leaving a one-line comment explaining the intentional retention so a future reader doesn't 'fix' it the wrong way.

src/components/ConnectionManager.tsx
```diff
@@ -18,10 +20,11 @@ export interface ConnectionForm {
...
   // safeStorage and reloaded the next time this connection is opened.
   savePassword: boolean;
-  // Opaque pass-through for mongodb-only entries — carried from a saved entry
-  // so re-save round-trips it. Not editable in this form; #113 will replace
-  // this with a real mongodb form and remove this field.
+  // Only meaningful when type === 'mongodb' AND mongoMode === 'uri'.
   connectionString?: string;
+  // UI state: which input mode the mongodb form is in. Persisted alongside
+  // the connection itself (derived on load: presence of connectionString).
+  mongoMode?: 'fields' | 'uri';
```
eugene-mobile, line 27: Style/follow-up only — the comment on `savedConnections.ts:13-15` already promised `#113 will tighten this with a discriminated union`. With this PR, the form has *two* mongodb-only fields (`connectionString`, `mongoMode`), and `mongoMode` is also load-bearing for the savedConnections invariant on round-trip. A discriminated union (e.g. `{ type: 'mysql' | 'postgres', ...common }` | `{ type: 'mongodb', mongoMode: 'fields' | 'uri', connectionString?: string, ...common }`) would let TS catch e.g. accidentally reading `mongoMode` from a postgres form, and would obviate `formFromSaved`'s `mongoMode` default for non-mongodb types. Not for this PR; flagging because the savedConnections comment promised it for #113 specifically.

eugene-mobile (commented):
Looks good overall — the dual-carry cleanup from #120 is fully gone, `formFromSaved` is a clean replacement, and `buildConnectBody()` correctly whitelists fields so neither `mongoMode` nor any other UI-only state leaks across the wire. Three non-blocking nits inline; one extra concern that doesn't fit on a changed line:

**Autocomplete display — `mongodb+srv://` URIs**

`src/components/ConnectionManager.tsx:351` (not touched by this PR but newly relevant) does `new URL(c.connectionString!).host` to derive the dropdown subtitle. WHATWG `URL` doesn't populate `host` consistently for non-special schemes, so for `mongodb+srv://...` the parser returns an empty `host` and the row falls through to displaying the raw URI — which often embeds the password. Compare with the server-side `connectionLabel` in `server/src/routes/connect.ts` which explicitly rewrites the scheme to `mongodb://` before parsing. Worth a follow-up to mirror that normalization here, and prefer `url.hostname` over `url.host` so we don't leak an explicit port or credentials.

src/components/ConnectionManager.tsx
eugene-mobile: Fixed in 92837c2a459b5ba5300f9756a6f4eed0298c9ff3. `buildConnectBody` is now presence-based: in URI mode the wire payload is just `{ type, database, ssl, connectionString }` — host/port/user/password are omitted entirely, not zeroed. `buildSubmitForm` no longer needs to overwrite the credentials; it just returns `form` unchanged in URI mode. The route's collision check and the `Number(port) || defaultPort` fallback are no longer load-bearing for correctness here.

src/api.ts
eugene-mobile: Fixed in 92837c2a459b5ba5300f9756a6f4eed0298c9ff3 along the lines you suggested — went with an early return in the connectionString branch so the URI-mode shape is unambiguous in the source rather than relying on a chain of conditional spreads. Same wire effect: `{ type, database, ssl, connectionString }` only, no `port: 0`.

src/components/ConnectionManager.tsx
eugene-mobile: Cleared in 92837c2a459b5ba5300f9756a6f4eed0298c9ff3 — `setDbType` now also drops `connectionString` when switching away from mongodb (`next === 'mongodb' ? p.connectionString : undefined`), matching the existing `mongoMode` reset. Smoke-tested mongodb→mysql in the browser; the next submit's payload no longer contains `connectionString`.

src/components/ConnectionManager.tsx
eugene-mobile: Agreed and deferred per your note. Filed as #125 with the proposed shape and the rationale (TS would catch `mongoMode` reads on non-mongodb forms and obviate the `formFromSaved` default). Scope spans `savedConnections.ts`, `ConnectionManager.tsx`, `api.ts`'s `ConnectFormInput`, and `App.tsx` callsites — too broad to wedge into this PR.

eugene-mobile:
Re: \"Autocomplete display — \`mongodb+srv://\` URIs\" in your review body —

Fixed in 92837c2a459b5ba5300f9756a6f4eed0298c9ff3. Added a small `hostFromConnectionString` helper in `ConnectionManager.tsx` that mirrors the server-side `connectionLabel` in `server/src/routes/connect.ts`: rewrites `mongodb+srv://` → `mongodb://` so WHATWG URL populates the host, prefers `url.hostname` over `url.host` so any explicit port is dropped, and on parse failure returns the literal `<connectionString>` placeholder rather than the raw URI (so a malformed entry can't leak credentials either).

Smoke-tested:
- `mongodb+srv://alice:s3cret@cluster.example.net/mydb` → autocomplete row shows `cluster.example.net · mydb` (password fully suppressed).
- Malformed (`not-a-url`) → shows `<connectionString>` instead of the raw string.

I left the helper local to the component rather than extracting a shared module across the front/back boundary — added a comment pointing at `connectionLabel` so the two stay in mental sync. If we end up needing this in a third place I'll lift it out.

PR #126, opened 2026-04-30
feat: QueryEditor MQL mode for MongoDB (#114)

### LINKED ISSUE #114: QueryEditor: switch to JSON/MQL mode

eugene-mobile:
Closes #114.

Switches the query editor into MQL mode whenever the active driver reports `queryMode === 'mql'`. SQL behaviour for MySQL/Postgres is untouched.

## What changed

**Backend**
- `/api/connect` and `/api/connect/status` now include `queryMode` from the active driver (`'sql' | 'mql' | null`).

**Frontend**
- `App` stores `queryMode` from the connect response and passes it to `QueryEditor`.
- `App.handleRun` parses the editor text as JSON when in MQL mode and POSTs `{ mql, schema }` via the new `api.queryMql` helper. SQL connections still POST `{ sql, schema }`.
- `App.handleTableSelect` seeds new tabs with a `find` snippet (instead of `SELECT *`) for MongoDB connections.
- `QueryEditor`:
  - Renders an MQL helper bar with a JSON example and a collection picker (populated from the schema).
  - The collection picker pre-fills the editor with `{ collection, operation: "find", filter: {}, limit: 100 }`.
  - The Format button rebinds to JSON pretty-print in MQL mode and its tooltip flips to "Format JSON".
  - Run validates as JSON before firing; parse errors show in the existing inline error strip and the request is not sent.
  - SQL-only behaviours (large-table warning, runtime-error line highlight) are gated on SQL mode.

## Acceptance criteria
- [x] Editor switches to JSON syntax/handling (not SQL formatting) when queryMode is `mql`.
- [x] Helper text above the editor shows an MQL example.
- [x] Collection selector dropdown populated from the current schema's collections.
- [x] Submitting sends a parsed MQL object to `/api/query` (not a string).
- [x] SQL behaviour for MySQL/Postgres is unchanged.

## Test plan
Browser smoke against an ephemeral local mongo (port 27017, db `smoke`, collections `users` + `orders`):
- [x] Connecting to mongodb returns `queryMode: "mql"` from `/api/connect`.
- [x] After connect: helper bar with `{ "collection": "users", "operation": "find", "filter": {} }` example is visible; collection dropdown lists `users`, `orders`.
- [x] Editing to invalid JSON and clicking Run shows an inline error and fires no `/api/query` request (verified via DevTools network inspection).
- [x] Valid `{ collection: "users", operation: "find", filter: { age: { $gt: 25 } } }` returns 200; UI shows Alice + Carol (age 30 + 41), excludes Bob (age 24).
- [x] `cd server && npm run build` clean; `npm run build` (frontend) clean; `cd server && npm test` 99 passing.

Note: the mongo result rendering is intentionally raw — `ResultsGrid` flattening for MongoDB documents is downstream in #115.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -219,8 +219,13 @@ export default function App() {
     if (queryMode === 'mql') {
       try {
         mqlPayload = JSON.parse(text);
-      } catch {
-        // Editor surfaces the parse error inline; bail without firing a request.
+      } catch (err) {
+        // Surface the error so callers that bypass QueryEditor.handleRunClick still see it.
+        setQueryError(err instanceof Error ? err.message : String(err));
+        return;
+      }
+      if (typeof mqlPayload !== 'object' || mqlPayload === null || Array.isArray(mqlPayload)) {
+        setQueryError('MQL request must be a JSON object (e.g. { "collection": "...", "operation": "find" }).');
         return;
       }
     }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,7 +8,7 @@ import { QueryEditor } from './components/QueryEditor';
 import { ResultsTable, type QueryResults } from './components/ResultsTable';
 import { ConnectionManager, type ConnectionForm } from './components/ConnectionManager';
 import { api } from './api';
-import type { SchemaData } from './api';
+import type { SchemaData, QueryMode } from './api';
 import { saveConnection } from './savedConnections';
 import { electronAPI } from './electronAPI';
 import { addHistoryEntry, listHistory, deleteHistoryEntry, clearHistory, type HistoryEntry } from './queryHistory';
@@ -46,6 +46,7 @@ export default function App() {
   }, [themeName]);
 
   const [connected, setConnected] = useState(false);
+  const [queryMode, setQueryMode] = useState<QueryMode>('sql');
   const [showConnectionModal, setShowConnectionModal] = useState(true);
   const [isConnecting, setIsConnecting] = useState(false);
   const [connectionError, setConnectionError] = useState<string | null>(null);
@@ -100,6 +101,7 @@ export default function App() {
       const friendly = form.name.trim();
       setConnectionName(friendly || res.connectionName);
       setConnectionHost(res.connectionName);
+      setQueryMode(res.queryMode);
       setHistory(listHistory(res.connectionName));
       setSavedQueries(listSavedQueries(res.connectionName));
       setSchemas(list);
@@ -142,6 +144,7 @@ export default function App() {
   const handleDisconnect = async () => {
     try { await api.disconnect(); } catch { /* ignore */ }
     setConnected(false);
+    setQueryMode('sql');
     setShowConnectionModal(true);
     setConnectionName('Not connected');
     setConnectionHost(null);
@@ -209,8 +212,18 @@ export default function App() {
 
   const handleRun = async () => {
     if (isRunning) return;
-    const sql = currentTab?.query?.trim();
-    if (!sql) return;
+    const text = currentTab?.query?.trim();
+    if (!text) return;
+
+    let mqlPayload: unknown = null;
+    if (queryMode === 'mql') {
+      try {
+        mqlPayload = JSON.parse(text);
+      } catch {
+        // Editor surfaces the parse error inline; bail without firing a request.
+        return;
+      }
+    }
 
     setIsRunning(true);
     setResults(null);
@@ -219,12 +232,14 @@ export default function App() {
 
     const started = Date.now();
     try {
-      const res = await api.query(sql, activeSchema);
+      const res = queryMode === 'mql'
+        ? await api.queryMql(mqlPayload, activeSchema)
+        : await api.query(text, activeSchema);
       setResults({ columns: res.columns, columnMeta: res.columnMeta, rows: res.rows });
       setExecTime(res.executionTime);
       if (connectionHost) {
         const entry = addHistoryEntry(connectionHost, {
-          sql, schema: activeSchema, executedAt: started,
+          sql: text, schema: activeSchema, executedAt: started,
           durationMs: res.executionTime, status: 'ok', rowCount: res.rows.length,
         });
         if (entry) setHistory(prev => [entry, ...prev].slice(0, 100));
@@ -234,7 +249,7 @@ export default function App() {
       setQueryError(message);
       if (connectionHost) {
         const entry = addHistoryEntry(connectionHost, {
-          sql, schema: activeSchema, executedAt: started,
+          sql: text, schema: activeSchema, executedAt: started,
           durationMs: Date.now() - started, status: 'error', error: message,
         });
         if (entry) setHistory(prev => [entry, ...prev].slice(0, 100));
@@ -306,7 +321,12 @@ export default function App() {
 
   const handleTableSelect = (name: string) => {
     setActiveTable(name);
-    addTab(`${name}.sql`, `SELECT *\nFROM \`${name}\`\nLIMIT 100;`);
+    if (queryMode === 'mql') {
+      const body = JSON.stringify({ collection: name, operation: 'find', filter: {}, limit: 100 }, null, 2);
+      addTab(`${name}.json`, body);
+    } else {
+      addTab(`${name}.sql`, `SELECT *\nFROM \`${name}\`\nLIMIT 100;`);
+    }
   };
 
   const switchTab = (id: string) => {
@@ -393,6 +413,7 @@ export default function App() {
               onChange={handleQueryChange}
               onRun={handleRun}
               isRunning={isRunning}
+              queryMode={queryMode}
               activeSchema={activeSchema}
               schemaData={schemaData}
               runtimeError={queryError}
--- a/src/api.ts
+++ b/src/api.ts
@@ -90,9 +90,11 @@ async function request<T>(path: string, init?: RequestInit): Promise<T> {
   return body;
 }
 
+export type QueryMode = 'sql' | 'mql';
+
 export const api = {
   connect(form: ConnectFormInput) {
-    return request<{ ok: boolean; connectionName: string }>('/api/connect', {
+    return request<{ ok: boolean; connectionName: string; queryMode: QueryMode }>('/api/connect', {
       method: 'POST',
       body: JSON.stringify(buildConnectBody(form)),
     });
@@ -110,7 +112,7 @@ export const api = {
   },
 
   status() {
-    return request<{ connected: boolean; connectionName: string | null }>('/api/connect/status');
+    return request<{ connected: boolean; connectionName: string | null; queryMode: QueryMode | null }>('/api/connect/status');
   },
 
   schemas() {
@@ -130,6 +132,13 @@ export const api = {
     return request<QueryResults & { columnMeta?: ColumnMeta[]; executionTime: number; affectedRows?: number; insertId?: number }>(
       '/api/query',
       { method: 'POST', body: JSON.stringify({ sql, schema }) }
+    );
+  },
+
+  queryMql(mql: unknown, schema: string) {
+    return request<QueryResults & { columnMeta?: ColumnMeta[]; executionTime: number; affectedRows?: number; insertId?: number }>(
+      '/api/query',
+      { method: 'POST', body: JSON.stringify({ mql, schema }) }
     );
   },
 
--- a/src/components/QueryEditor.tsx
+++ b/src/components/QueryEditor.tsx
@@ -156,7 +156,7 @@ export function QueryEditor({
 
   const now = Date.now();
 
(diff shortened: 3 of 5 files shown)


eugene-mobile (commented):
Solid keystone PR — backend plumbing is clean and the gate on `isMql` is consistently applied (large-table warning, runtime-error line highlight). A few real UX bugs around the parse-error strip and the collection picker are worth fixing before merge; the rest are non-blocking.

src/components/QueryEditor.tsx
```diff
@@ -597,6 +624,50 @@ export function QueryEditor({
...
+              <select
+                value=""
+                onChange={(e) => {
+                  const name = e.target.value;
+                  if (!name) return;
+                  const snippet = JSON.stringify(
+                    { collection: name, operation: 'find', filter: {}, limit: 100 },
+                    null, 2,
+                  );
+                  onChange(snippet);
```
eugene-mobile: **Blocking-ish UX bug — collection picker clobbers editor content.**

`onChange(snippet)` unconditionally replaces the entire editor value. If the user has been editing a query and accidentally picks from the dropdown (or wants to insert a *second* collection into a multi-step query), their work disappears with no confirmation and no native undo (the textarea's undo stack is broken because the value is round-tripped through React state).

Suggest one of:
- Only replace when `value.trim() === ''`; otherwise insert at caret or no-op with a tooltip.
- Rename the label to "Replace with…" and confirm if non-empty.
- Insert as a new tab via `addTab` (matches `handleTableSelect`).

The label currently reads "Insert collection:" which is misleading — it replaces, not inserts.

src/components/QueryEditor.tsx
```diff
@@ -232,18 +246,31 @@ export function QueryEditor({
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [runtimeError]);
 
+  const handleRunClick = () => {
+    if (isMql && value.trim()) {
+      try {
+        JSON.parse(value);
+        setFormatError(null);
+      } catch (err) {
+        setFormatError(err instanceof Error ? err.message : String(err));
```
eugene-mobile: **Misleading error label.** When JSON.parse fails on Run, this writes the parse error into `formatError`, which is then rendered (line 608) with the hardcoded prefix `Format failed: …`. The user clicked **Run**, not Format. Either:
- Introduce a separate `runError` state for the MQL parse-on-run path, or
- Make the strip prefix conditional on what triggered it (`Invalid JSON: …` when called from `handleRunClick`, `Format failed: …` from `handleFormat`).

Minor, but it'll confuse users who never touched the format button.

src/components/QueryEditor.tsx
```diff
@@ -232,18 +246,31 @@ export function QueryEditor({
...
     selectErrorLine(loc.lineStart, loc.lineEnd);
     // Intentionally excluding `value` from deps: only react to a fresh runtime error.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [runtimeError]);
 
+  const handleRunClick = () => {
+    if (isMql && value.trim()) {
+      try {
+        JSON.parse(value);
+        setFormatError(null);
```
eugene-mobile: **Stale parse-error banner.** When the user fixes their JSON after a failed Run/Format, the red strip persists until they click Run again or hit the × dismiss. The textarea onChange in this file (line ~680, unchanged) does not clear formatError. Consider clearing it from the parent value-onChange handler in MQL mode, e.g. wrap the editor onChange to also call setFormatError(null) on first keystroke after an error.

src/App.tsx
```diff
@@ -209,8 +212,18 @@ export default function App() {
...
+    const text = currentTab?.query?.trim();
+    if (!text) return;
+
+    let mqlPayload: unknown = null;
+    if (queryMode === 'mql') {
+      try {
+        mqlPayload = JSON.parse(text);
+      } catch {
+        // Editor surfaces the parse error inline; bail without firing a request.
+        return;
```
eugene-mobile, line 229: **Silent bail on parse failure.** When `JSON.parse(text)` throws here, `handleRun` returns without setting `queryError`, without resetting `isRunning` (it isn't set yet, so OK), and — crucially — without any side effect the user can observe from this component. The visible inline error currently relies on the user having clicked the toolbar Run button (which routes through `QueryEditor.handleRunClick` and sets `formatError` first).

This is fine *today* because every entry point goes through `handleRunClick` first, but it's fragile: if a future caller invokes `handleRun` directly (programmatic run, keyboard shortcut at App level, command palette), the failure will be completely silent. Consider surfacing the parse error here too via `setQueryError(`Invalid MQL JSON: ${err.message}`)`.

src/App.tsx
```diff
@@ -100,6 +101,7 @@ export default function App() {
       const friendly = form.name.trim();
       setConnectionName(friendly || res.connectionName);
       setConnectionHost(res.connectionName);
+      setQueryMode(res.queryMode);
```
eugene-mobile, line 104: **Type narrowing — `res.queryMode` is typed `QueryMode` (non-null) on `/api/connect`, but `QueryMode | null` on `/api/status`.** Currently no code path seeds `queryMode` from `api.status()` (e.g. on app reload while still connected to an Electron-persisted session), so this is fine for now.

If you ever wire a status-driven reconnect, `setQueryMode(QueryMode | null)` will be a TS error against `useState<QueryMode>('sql')`. Worth widening the state to `QueryMode | null` proactively, or coercing `res.queryMode ?? 'sql'` at the boundary.

src/api.ts
```diff
@@ -133,6 +135,13 @@ export const api = {
     );
   },
 
+  queryMql(mql: unknown, schema: string) {
```
eugene-mobile, line 138: **`mql: unknown` is too loose at the boundary.** `JSON.parse('null')`, `JSON.parse('42')`, and `JSON.parse('"hello"')` all succeed in `App.handleRun`, and any of those will be POSTed to `/api/query`. The server should (and presumably does) reject non-objects, but a client-side guard like `typeof mql === 'object' && mql !== null && !Array.isArray(mql)` before firing would give the user a faster, friendlier inline error. Non-blocking — server validation is the source of truth.

src/App.tsx
```diff
@@ -142,6 +144,7 @@ export default function App() {
   const handleDisconnect = async () => {
     try { await api.disconnect(); } catch { /* ignore */ }
     setConnected(false);
+    setQueryMode('sql');
```
eugene-mobile, line 147: **Resetting `queryMode` to `'sql'` on disconnect is fine but not load-bearing** — the editor is hidden behind the connect modal at that point. If you later allow opening the connect modal mid-session without disconnecting first, this reset wouldn't fire and a stale `'mql'` could leak into a SQL connection. Documenting the assumption with a comment, or using `null` as the disconnected sentinel and deriving `isMql = queryMode === 'mql'` everywhere, would be more defensive.

src/components/QueryEditor.tsx
eugene-mobile: Fixed in 938b3ca. The picker now inserts the snippet at `textareaRef.current`'s `selectionStart`/`selectionEnd` when the editor is non-empty, and only replaces wholesale when `value.trim() === ''`. Cursor moves to the end of the inserted snippet.

src/components/QueryEditor.tsx
eugene-mobile: Fixed in 938b3ca by renaming `formatError` to `editorError` typed as `{ kind: 'format' | 'run', message }`. The banner prefix is now derived from `kind`, so Run-time JSON failures render as "Parse failed: …" and toolbar Format failures stay "Format failed: …".

src/components/QueryEditor.tsx
eugene-mobile: Fixed in 938b3ca. The textarea `onChange` now clears `editorError` on the first keystroke after an error, regardless of whether it came from Run or Format.

src/App.tsx
eugene-mobile: Fixed in 938b3ca. `App.handleRun` now calls `setQueryError(...)` in the catch path of the JSON.parse so callers that bypass `QueryEditor.handleRunClick` still surface an inline error in the ResultsTable banner.

src/api.ts
eugene-mobile: Addressed in 938b3ca. Added a client-side guard in `App.handleRun` after the JSON.parse: if the parsed value isn't a plain object (`null`, array, or scalar), we surface an inline error via `setQueryError` and skip the POST. Server validation remains the source of truth.

src/App.tsx
eugene-mobile: Deferring per triage — YAGNI for now. No code path currently seeds `queryMode` from `api.status()`, so the `QueryMode` (non-null) state is fine. I'll widen to `QueryMode | null` when status-driven reseed gets wired up.

src/App.tsx
eugene-mobile: Deferring — touching every `isMql === 'mql'` site to switch to a `null` sentinel for no current bug isn't worth the churn. The modal-on-disconnect invariant is still in place; if we ever allow opening connect mid-session, I'll revisit by switching the disconnected state to `null` and deriving `isMql` everywhere.

eugene-mobile (commented):
Re-review of 938b3ca: 5/5 actioned items LGTM, 2 deferrals acknowledged. Approving with two minor non-blocking notes captured inline.

src/App.tsx
```diff
@@ -209,8 +212,23 @@ export default function App() {
...
+    const text = currentTab?.query?.trim();
+    if (!text) return;
+
+    let mqlPayload: unknown = null;
+    if (queryMode === 'mql') {
+      try {
+        mqlPayload = JSON.parse(text);
+      } catch (err) {
+        // Surface the error so callers that bypass QueryEditor.handleRunClick still see it.
+        setQueryError(err instanceof Error ? err.message : String(err));
```
eugene-mobile, line 224: Non-blocking: this catch branch is effectively dead code in the current call graph. `App.handleRun` is only invoked via `QueryEditor.onRun`, and `handleRunClick` already JSON.parses and bails on failure before calling `onRun()`. The defensive `setQueryError` here is fine as a safety net (worth keeping for future callers, e.g. the deferred status-driven seed), but the duplicated parse is worth a comment so a future reader doesn't "simplify" by removing one side.

Also: the object-shape branch below renders the error in `ResultsTable` (via `queryError`) rather than inline in the editor banner like the JSON parse error does in `handleRunClick`. The two failure modes for the same Run click thus surface in two different places. Consider routing the shape error through the editor banner too (or move the JSON.parse + shape check into `handleRunClick` and pass the parsed payload through `onRun(payload)`).

src/components/QueryEditor.tsx
```diff
@@ -597,6 +625,67 @@ export function QueryEditor({
...
+                  const snippet = JSON.stringify(
+                    { collection: name, operation: 'find', filter: {}, limit: 100 },
+                    null, 2,
+                  );
+                  if (!value.trim()) {
+                    onChange(snippet);
+                  } else {
+                    const el = textareaRef.current;
+                    const start = el?.selectionStart ?? value.length;
+                    const end = el?.selectionEnd ?? value.length;
```
eugene-mobile, line 660: Non-blocking: `selectionStart` / `selectionEnd` on an unfocused textarea reflect the last cursor position (or 0 if it's never been focused). The `?? value.length` fallback only fires when `textareaRef.current` itself is null, so a never-focused textarea will insert at offset 0, not at end. In practice the user has almost certainly clicked into the editor before reaching for the collection picker, so this is a theoretical edge case — but if you want it bulletproof, gate on `el && document.activeElement === el` and otherwise append to end.

src/components/QueryEditor.tsx
```diff
@@ -597,6 +625,67 @@ export function QueryEditor({
...
+                    const next = value.substring(0, start) + snippet + value.substring(end);
+                    onChange(next);
```
eugene-mobile, line 662: Minor: the collection-picker path calls `onChange(...)` programmatically, which does not trigger the textarea's `onChange` keystroke-clear. So if there's a stale `kind: 'run'` parse error, picking a collection (which inserts valid JSON) leaves the banner up. Cheap fix: `if (editorError) setEditorError(null);` in this handler too. Not blocking; the next keystroke clears it.

PR #127, opened 2026-05-01
feat: flatten nested MongoDB documents in ResultsGrid (#115)

### LINKED ISSUE #115: ResultsGrid: flatten nested documents one level

eugene-mobile:
Closes #115

## Summary
- Detects nested objects/arrays in result rows and one-level flattens them: `{ address: { city, zip } }` becomes columns `address.city` / `address.zip`, replacing the original column.
- Deeper objects and arrays render as a truncated JSON preview with a hint icon; clicking the cell opens a modal with the pretty-printed value plus a "Copy JSON" button.
- Detection is gated on \"any non-null object value in the rows\", so SQL drivers (which only emit primitives) bypass the entire pipeline — MySQL/Postgres rendering is identical to before.

## Acceptance criteria
- [x] Nested objects/arrays in result rows are detected
- [x] One level of nesting flattened by default — `address.city`, `address.zip`
- [x] Deeper nesting renders as JSON string with an expand affordance (modal)
- [x] No regression for MySQL/Postgres rows

## Design choices
- **Replace, don't duplicate**: when a column expands, the parent column is dropped — showing both `address` and `address.city` in the same grid is just clutter. Deeper nested children that aren't flattened still appear (as truncated JSON in the synthetic child cell).
- **Modal over inline expansion**: a modal keeps the grid layout intact, gives the user enough room for pretty-printed JSON, and reuses the existing Esc-to-close pattern from the Update / Delete confirmations. A click on any complex cell opens it; the cell's `cursor: zoom-in` signals the affordance.
- **`_id` stays a string**: the mongo driver already serialises ObjectId via `toHexString()`, so `_id` lands as a primitive 24-char string and renders the same as any other id.
- **Export reflects what you see**: CSV/JSON export now uses the flattened columns/rows; complex cells JSON-stringify so spreadsheets get something readable.

## Test plan
- [x] `npm run build` clean (root + server)
- Manual smoke (mongo 7 in docker, with the 4 seeded docs from issue):
  - **users / find** showed columns `_ID`, `NAME`, `ADDRESS.CITY`, `ADDRESS.ZIP`, `TAGS`. The `address` column was replaced by `address.city` (Toronto, Vancouver) and `address.zip` (M5V, V6B). `tags` rendered as `["admin","eng"]` / `["user"]` strings — no `[object Object]` anywhere. Clicking a `tags` cell opened the modal showing `[\n  \"admin\",\n  \"eng\"\n]` with header `tags · array · 2 items`.
  - **events / find** showed columns `_ID`, `PAYLOAD.META`. The `PAYLOAD.META` cell rendered `{\"source\":\"web\",\"deep\":{\"nested\":\"valu…` truncated with a hint icon. Clicking opened the modal pretty-printing the full nested object including the `deep.nested` level.
- SQL regression (postgres 16 in docker): `SELECT 1 AS x, 'hello' AS y` rendered with columns `X`, `Y` and a single row `1`, `hello`. Cell cursors all default — no `zoom-in` hints, no flattening logic engaged.

### DIFF
--- a/src/components/ResultsTable.tsx
+++ b/src/components/ResultsTable.tsx
@@ -5,12 +5,24 @@ import type { ColumnMeta, SchemaData } from '../api';
 import { InsertRowDialog } from './InsertRowDialog';
 import { rowsToCsv, rowsToJson, downloadBlob, sanitizeFilename } from '../export';
 
+// MongoDB documents arrive with nested objects/arrays in cell values.
+// Other drivers always emit primitives. We type the row map permissively so
+// the same component handles both; nested values are flattened or serialised
+// for display (see `flattenRowsForDisplay`).
+export type RawCellValue = string | number | boolean | null | unknown[] | Record<string, unknown>;
+
 export interface QueryResults {
   columns: string[];
   columnMeta?: ColumnMeta[];
-  rows: Record<string, string | number | boolean | null>[];
+  rows: Record<string, RawCellValue>[];
 }
 
+// `Row` carries primitive cell values only — that is the contract with the
+// editing/CRUD callbacks (onUpdateCell / onDeleteRow / onInsertRow). For
+// MongoDB results that contain nested objects/arrays we work internally with
+// the wider `DisplayRow` (see `flattenRowsForDisplay`); editing is gated on
+// `meta.orgTable && meta.orgName`, which Mongo leaves empty, so a non-primitive
+// cell never reaches the CRUD callbacks.
 type Row = Record<string, string | number | boolean | null>;
 type CellValue = string | number | boolean | null;
 
@@ -81,10 +93,125 @@ function isBoolColumn(schemaData: SchemaData | undefined, meta: ColumnMeta | und
   return col?.type === 'tinyint(1)';
 }
 
-function cellDisplayValue(val: string | number | boolean | null, meta: ColumnMeta | undefined, schemaData: SchemaData | undefined): string {
-  if (val === null) return '';
+function cellDisplayValue(val: RawCellValue, meta: ColumnMeta | undefined, schemaData: SchemaData | undefined): string {
+  if (val === null || val === undefined) return '';
   if (isBoolColumn(schemaData, meta)) return val === 1 || val === true ? 'true' : 'false';
+  if (typeof val === 'object') {
+    try { return JSON.stringify(val); } catch { return String(val); }
+  }
   return String(val);
+}
+
+// Nested-document flattening (MongoDB).
+// SQL drivers always emit primitive cell values, so `hasNesting` returns false
+// and the original rows/columns are used unchanged — there is no behavioural
+// change for MySQL/Postgres results.
+//
+// One level of plain-object nesting is flattened: `{ address: { city, zip } }`
+// becomes columns `address.city` and `address.zip`, replacing `address`.
+// Deeper objects and arrays remain as raw values in the flattened row and are
+// rendered as truncated JSON with a click-to-expand modal.
+function isPlainObject(v: unknown): v is Record<string, unknown> {
+  return typeof v === 'object' && v !== null && !Array.isArray(v);
+}
+
+function hasNesting(rows: Record<string, RawCellValue>[]): boolean {
+  for (const r of rows) {
+    for (const k in r) {
+      const v = r[k];
+      if (v !== null && typeof v === 'object') return true;
+    }
+  }
+  return false;
+}
+
+interface FlattenedResult {
+  columns: string[];
+  rows: Record<string, RawCellValue>[];
+}
+
+function flattenRowsForDisplay(
+  columns: string[],
+  rows: Record<string, RawCellValue>[],
+): FlattenedResult {
+  // First pass: decide which top-level columns hold an object on at least one
+  // row. Those columns get expanded to `parent.child` columns. Arrays stay put
+  // (heterogeneous shapes don't flatten cleanly), as do columns whose values
+  // are always primitive.
+  const expandTargets = new Set<string>();
+  const childKeyOrder = new Map<string, string[]>();
+  for (const col of columns) {
+    const seenChildren: string[] = [];
+    let everObject = false;
+    for (const r of rows) {
+      const v = r[col];
+      if (isPlainObject(v)) {
+        everObject = true;
+        for (const k of Object.keys(v)) {
+          if (!seenChildren.includes(k)) seenChildren.push(k);
+        }
+      }
+    }
+    if (everObject) {
+      expandTargets.add(col);
+      childKeyOrder.set(col, seenChildren);
+    }
+  }
+
+  if (expandTargets.size === 0) return { columns, rows };
+
+  const newColumns: string[] = [];
+  for (const col of columns) {
+    if (expandTargets.has(col)) {
+      for (const child of childKeyOrder.get(col)!) newColumns.push(`${col}.${child}`);
+    } else {
+      newColumns.push(col);
+    }
+  }
+
+  const newRows: Record<string, RawCellValue>[] = rows.map(r => {
+    const out: Record<string, RawCellValue> = {};
+    for (const col of columns) {
+      if (expandTargets.has(col)) {
+        const v = r[col];
+        const children = childKeyOrder.get(col)!;
+        if (isPlainObject(v)) {
+          for (const child of children) {
+            out[`${col}.${child}`] = (v[child] ?? null) as RawCellValue;
+          }
+        } else {
+          // Row didn't have an object at this column — fill children with null
+          // (or carry the non-object value into the first child if it exists).
+          for (const child of children) out[`${col}.${child}`] = null;
+          if (v !== undefined && v !== null) {
+            // Preserve the unexpected value on the first synthetic child so
+            // it isn't silently dropped from the grid.
+            out[`${col}.${children[0]}`] = v;
+          }
+        }
+      } else {
+        out[col] = r[col] ?? null;
+      }
+    }
+    return out;
+  });
+
+  return { columns: newColumns, rows: newRows };
+}
+
+function isComplexValue(v: RawCellValue): v is unknown[] | Record<string, unknown> {
+  return v !== null && typeof v === 'object';
+}
+
+function jsonPreview(v: unknown, max: number): string {
+  let s: string;
+  try {
+    s = JSON.stringify(v);
+  } catch {
+    s = String(v);
+  }
+  if (s.length <= max) return s;
+  return s.slice(0, max - 1) + '…';
 }
 
 // Long values blow out column widths and force users to scroll. We truncate
@@ -190,6 +317,8 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
   const [filterOpen, setFilterOpen] = useState(false);
   const [filterText, setFilterText] = useState('');
   const filterInputRef = useRef<HTMLInputElement | null>(null);
+  // Full-cell viewer for nested objects/arrays from MongoDB documents.
+  const [expandedCell, setExpandedCell] = useState<{ col: string; value: RawCellValue } | null>(null);
 
   // Column layout: order and widths
   const [colOrder, setColOrder] = useState<string[]>([]);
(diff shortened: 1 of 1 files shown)


eugene-mobile (commented):
Approve with non-blocking follow-ups. The flatten + modal flow looks correct and is correctly gated by `hasNesting` so SQL is untouched. Concerns are mostly around perf (no memoisation), the heterogeneous-shape fallback, deeper-nested BSON types arriving from the driver, and minor UX/consistency nits in the modal.

src/components/ResultsTable.tsx
```diff
@@ -288,10 +427,19 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
...
   };
 
+  // Result set after one-level flattening of nested documents (no-op for SQL).
+  // Memoised by `results` identity since the parent always replaces `results`
+  // wholesale on a new query.
+  const flattened = (() => {
+    if (!results) return null;
+    if (!hasNesting(results.rows)) return { columns: results.columns, rows: results.rows };
+    return flattenRowsForDisplay(results.columns, results.rows);
+  })();
```
eugene-mobile: Non-blocking, but the comment above claims this is `Memoised by results identity` and it isn't — this is a plain IIFE that re-runs on every render. With 10k Mongo rows you'll re-walk the whole result set (in `hasNesting` and again in `flattenRowsForDisplay`) on every keystroke in the filter input, every selection change, every resize. Wrap in `useMemo(() => ..., [results])` to actually match the comment. While here, `hasNesting` itself is also worth memoising independently — for SQL it short-circuits cheaply, but for any Mongo result it scans every cell of every row before falling through to `flattenRowsForDisplay` which re-walks the same data.

src/components/ResultsTable.tsx
```diff
@@ -81,12 +93,127 @@ function isBoolColumn(schemaData: SchemaData | undefined, meta: ColumnMeta | und
...
+            out[`${col}.${child}`] = (v[child] ?? null) as RawCellValue;
+          }
+        } else {
+          // Row didn't have an object at this column — fill children with null
+          // (or carry the non-object value into the first child if it exists).
+          for (const child of children) out[`${col}.${child}`] = null;
+          if (v !== undefined && v !== null) {
+            // Preserve the unexpected value on the first synthetic child so
+            // it isn't silently dropped from the grid.
+            out[`${col}.${children[0]}`] = v;
```
eugene-mobile: Heterogeneous-shape fallback is silently misleading. If `address` is `{city, zip}` on row 1 but `"123 Main St"` (string) or `["a","b"]` (array) on row 2, the non-object value gets shoved into `out[\`address.${children[0]}\`]` — i.e. into the synthetic `address.city` column. A user reading the grid sees `"123 Main St"` under a column literally labelled `address.city` with no indication the source value wasn't an object. Suggest either (a) keep the parent column when shapes are mixed and don't expand at all, or (b) write the orphan value into a dedicated synthetic column whose name signals the shape mismatch (e.g. retain `address` alongside `address.city`/`address.zip`, with the expanded children null on rows where the parent isn't an object). Currently you both drop the parent column and corrupt the first child column.

src/components/ResultsTable.tsx
```diff
@@ -81,12 +93,127 @@ function isBoolColumn(schemaData: SchemaData | undefined, meta: ColumnMeta | und
...
+// change for MySQL/Postgres results.
+//
+// One level of plain-object nesting is flattened: `{ address: { city, zip } }`
+// becomes columns `address.city` and `address.zip`, replacing `address`.
+// Deeper objects and arrays remain as raw values in the flattened row and are
+// rendered as truncated JSON with a click-to-expand modal.
+function isPlainObject(v: unknown): v is Record<string, unknown> {
+  return typeof v === 'object' && v !== null && !Array.isArray(v);
+}
+
```
eugene-mobile, line 126: Two related concerns:

**Perf**: detection scans every cell of every row. For a 10k-row Mongo result with no nesting on any row it's O(rows*cols) per render (and this runs every render — see comment on line 437). You already bail on the first nested value found, but consider also capping the scan at the first N rows (say 200) — Mongo result-set shape is overwhelmingly homogeneous within a single query, and the cost of the rare false-negative (a deeper row with an unexpected object) is just a non-flattened cell, which the existing `ComplexCellPreview` handles fine.

**Driver-shaped values**: outside this PR's diff but worth flagging here since the flatten lives next door — `server/src/drivers/mongodb.ts#serializeValue` handles `ObjectId`, `Date`, `Buffer`, `bigint`, arrays, and plain objects, but not `Decimal128`, `Long`, `Binary`, `UUID`, or `Timestamp`. Those BSON types fall into the generic `typeof === 'object'` recursion and surface in the grid as `{$numberDecimal: "1.23"}` / `{buffer: {...}, position: ...}` shapes — which `hasNesting` will then helpfully expand into columns like `price.$numberDecimal`. Worth a follow-up so the next person inspecting a Mongo collection with monetary fields isn't surprised.

src/components/ResultsTable.tsx
```diff
@@ -765,48 +936,58 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
...
+                  const cellVal = row[col] ?? null;
+                  const complex = isComplexValue(cellVal);
                   return (
                     <td
                       key={col}
-                      onClick={(e) => { e.stopPropagation(); setSelectedRow(i); setSelectedCol(colIdx); tableWrapRef.current?.focus(); }}
+                      onClick={(e) => {
+                        e.stopPropagation();
+                        setSelectedRow(i); setSelectedCol(colIdx); tableWrapRef.current?.focus();
+                        if (complex) setExpandedCell({ col, value: cellVal });
```
eugene-mobile: Cell click now does three things — select cell, focus wrap, AND open modal for complex values. That's fine for the click affordance, but worth a thought: complex cells now have no way to be selected/focused without also opening the modal. If the user wants to Ctrl+C copy a complex cell's JSON without opening the modal (the keyboard `Ctrl+C` path on line 815 already does the right thing via `cellDisplayValue`), they have to use right-click → Copy. Consider gating the auto-open behind a small affordance on the cell (e.g. only opening when the click target is the zoom-in icon, or requiring a double-click), so single-click remains pure cell selection consistent with primitive cells.

src/components/ResultsTable.tsx
```diff
@@ -1132,10 +1315,89 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
...
+          >
+            <div style={{ display: 'flex', alignItems: 'center', gap: 10, marginBottom: 10 }}>
+              <h3 style={{ margin: 0, fontSize: 13, fontWeight: 600, fontFamily: '"JetBrains Mono", monospace' }}>{expandedCell.col}</h3>
+              <span style={{ fontSize: 11, color: t.textMuted }}>
+                {Array.isArray(expandedCell.value) ? `array · ${(expandedCell.value as unknown[]).length} item${(expandedCell.value as unknown[]).length === 1 ? '' : 's'}` : 'object'}
+              </span>
+              <div style={{ flex: 1 }} />
+              <button
+                onClick={() => {
+                  try { copyToClipboard(JSON.stringify(expandedCell.value, null, 2)); } catch { /* noop */ }
```
eugene-mobile, line 1356: `Copy JSON` here copies `JSON.stringify(value, null, 2)` (pretty-printed, 2-space indent), but `cellDisplayValue` (used by Ctrl+C and the context menu's Copy) emits compact `JSON.stringify(value)`. So the same complex value yields different clipboard contents depending on which copy affordance the user uses. Either is defensible, but pick one — pretty for both is probably the better UX since users opening the modal want a readable form anyway.

src/components/ResultsTable.tsx
```diff
@@ -275,6 +404,16 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
...
   }, [filterOpen]);
 
+  // Esc closes the expanded-cell viewer.
+  useEffect(() => {
+    if (!expandedCell) return;
+    const onKey = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') { e.preventDefault(); setExpandedCell(null); }
+    };
+    window.addEventListener('keydown', onKey);
+    return () => window.removeEventListener('keydown', onKey);
```
eugene-mobile, line 424: This Esc handler doesn't `stopImmediatePropagation`, and the existing tableWrap onKeyDown (line 810) and confirm-dismiss handler (line 387) also listen for Esc. Net effect: pressing Esc with the expand modal open will close the modal AND clear the cell selection (`setSelectedRow(null); setSelectedCol(null)` from line 811). Probably fine — most users won't notice — but worth a comment explaining the precedence is intentional, or guarding the tableWrap handler with `if (expandedCell) return` to keep selection sticky after the modal dismisses.

src/components/ResultsTable.tsx
eugene-mobile: Fixed in 0ef86c7cfe7abc8463757f21279e7ee1715310ea. Wrapped `flattened` in `useMemo` keyed on `results` identity and dropped the stale comment. `hasNesting` lives inside the same memo so it isn't re-run independently.

src/components/ResultsTable.tsx
eugene-mobile: Fixed in 0ef86c7cfe7abc8463757f21279e7ee1715310ea — option (a). The expand-target gate now requires *every* non-null occurrence at a given column to be a plain object. Mixed shapes leave the parent column intact and the object rows render via `ComplexCellPreview`.

src/components/ResultsTable.tsx
eugene-mobile: Filed as #128 for the BSON-types serialization gap (Decimal128/Long/Binary/UUID/Timestamp). Leaving the bounded-scan optimisation deferred — homogeneous Mongo result sets are the common case and the perf concern is the same root issue as the duplicate-walk, which is mitigated by the new `useMemo` in 0ef86c7cfe7abc8463757f21279e7ee1715310ea.

src/components/ResultsTable.tsx
eugene-mobile: Fixed in 0ef86c7cfe7abc8463757f21279e7ee1715310ea. Cell click is now pure selection. The zoom-in icon is a real `<button>` with `e.stopPropagation()` that's the only affordance for opening the modal, so Ctrl+C on a selected complex cell is now a clean path to copying without opening anything.

src/components/ResultsTable.tsx
eugene-mobile: Fixed in 0ef86c7cfe7abc8463757f21279e7ee1715310ea. `cellDisplayValue` now returns `JSON.stringify(val, null, 2)` for object/array values, so Ctrl+C and right-click Copy match the modal's "Copy JSON" output exactly. Verified end-to-end.

src/components/ResultsTable.tsx
eugene-mobile: Fixed in 0ef86c7cfe7abc8463757f21279e7ee1715310ea. The tableWrap Esc handler is now guarded with `if (expandedCell) break;` so closing the modal preserves the underlying cell selection. Verified in browser: select cell → open modal via icon → Esc → modal closes, selection stays.

src/components/ResultsTable.tsx
eugene-mobile: Follow-up regression fixed in d608e6a6ae99e4d7899352fa98afa0f40800beb3. The pretty-printed JSON is correct for single-cell copy but row-copy joins cells with tabs to build a TSV, so the embedded newlines split one cell across multiple spreadsheet rows on paste.

`cellDisplayValue` now takes an opt-in `multiline` flag (default true). Both row-copy callsites pass `false`, single-cell paths keep the pretty-printed default. Verified in browser: single-cell Copy on a tags array → `[\n  "admin",\n  "ops"\n]`; Copy row → `1\tAlice\tToronto\tM5V\t["admin","ops"]` (tab-separated, no embedded newlines).

PR #129, opened 2026-05-01
fix(mongodb): serialize remaining BSON scalar types (#128)

### LINKED ISSUE #128: fix(mongodb): serialize remaining BSON scalar types
## Background

`server/src/drivers/mongodb.ts#serializeValue` handles `ObjectId` and `Date` explicitly, but several other common BSON scalar types fall through into the generic recursive object branch. They end up as `{}` or `{ "$numberDecimal": "..." }`-shaped junk in the grid instead of meaningful primitives.

Spotted in the PR #127 review (https://github.com/appspace/helix/pull/127#pullrequestreview-4211424600).

## Affected types and current rendering

| Type | Current render | Expected |
|---|---|---|
| `Decimal128` | `{ "$numberDecimal": "..." }` (from `toJSON`) or `{}` | string of the numeric value |
| `Long` (Int64) | `{ "low": ..., "high": ..., ... }` | string (preserves precision past `Number.MAX_SAFE_INTEGER`) |
| `Binary` | `{}` | hex or base64 string with subtype hint |
| `UUID` (Binary subtype 4) | `{}` | canonical UUID string `xxxxxxxx-xxxx-...` |
| `Timestamp` | `{}` | `T(<seconds>, <increment>)` or ISO-like string |

## Fix shape

Extend the `switch` (or the early-return chain) in `serializeValue` to detect each `_bsontype` (or instanceof from the `bson` package) and produce a primitive string. Reuse the same approach already in place for `ObjectId`.

Add unit tests in `server/src/drivers/__tests__/mongodb.serialize.test.ts` (or wherever the existing serialize tests live) covering each type.

## Notes

- Type detection should be defensive — both `instanceof` and `_bsontype` checks handle the case where users pull in their own BSON build.
- `Binary` subtype 4 is UUID; preserve the UUID string form rather than the raw hex.

eugene-mobile:
Closes #128

## Summary

After PR #127 (ResultsGrid flatten), MongoDB documents containing `Decimal128`, `Long`, `Binary`, `UUID`, or `Timestamp` fields rendered as JSON-shaped junk (e.g. `{"$numberDecimal":"..."}`) — or, worse, the flatten gate could split them into bogus `field.$numberDecimal` columns.

`server/src/drivers/mongodb.ts#serializeValue` now detects each type via the bson `_bsontype` tag (duck-typed so a duplicate `bson` copy in the dep graph doesn't break `instanceof`) and emits a primitive string.

## Per-type rendering

| Type | Output | Rationale |
|---|---|---|
| `Decimal128` | `value.toString()` | exact numeric form, no precision loss |
| `Long` | `value.toString()` | preserves precision past `Number.MAX_SAFE_INTEGER` |
| `Binary` subtype 4 | canonical UUID via `toUUID().toString()` | matches Compass / mongosh display |
| `Binary` (other subtypes) | `Binary(<subType>,<base64>)` | base64 is compact and round-trippable; subtype is shown so users can tell `0` (generic) from `2` (deprecated) etc. |
| `Timestamp` | `Timestamp(t, i)` | matches mongosh's printed form; `t` (seconds since epoch) and `i` (increment) are the meaningful pair |

`Timestamp` is checked **before** `Long` because `Timestamp extends Long`.

## AC checklist

- [x] `Decimal128` renders as numeric string (positive, negative, zero)
- [x] `Long` renders as string (max int64 covered)
- [x] `Binary` subtype 4 renders as canonical UUID
- [x] Other `Binary` renders as readable string with subtype + base64
- [x] `Timestamp` renders as `Timestamp(t, i)`
- [x] Existing `ObjectId`/`Date`/`null`/`undefined` paths untouched
- [x] Nested arrays / sub-documents recurse correctly

## Test plan

- [x] `cd server && npm test` — 105/105 green (6 new cases under `MongoDBDriver.query – BSON scalar serialization`)
- [ ] Browser smoke: connect to a MongoDB instance, insert a doc containing all 5 BSON types, run `find {}` in MQL mode, verify each cell shows a readable string (not an object/JSON blob)

### DIFF
--- a/server/src/drivers/mongodb.test.ts
+++ b/server/src/drivers/mongodb.test.ts
@@ -273,6 +273,90 @@ describe('MongoDBDriver.query – find', () => {
   });
 });
 
+describe('MongoDBDriver.query – BSON scalar serialization', () => {
+  it('serializes Decimal128 to its numeric string (positive, negative, zero)', async () => {
+    const dec = (s: string) => ({ _bsontype: 'Decimal128', toString: () => s });
+    mockCursor.toArray.mockResolvedValueOnce([
+      { a: dec('123.45'), b: dec('-987654321.0000001'), c: dec('0') },
+    ]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'c', operation: 'find' }),
+    );
+    expect(r.rows[0]).toEqual({ a: '123.45', b: '-987654321.0000001', c: '0' });
+  });
+
+  it('serializes Long to its string form (preserves precision past Number.MAX_SAFE_INTEGER)', async () => {
+    const long = (s: string) => ({ _bsontype: 'Long', toString: () => s });
+    mockCursor.toArray.mockResolvedValueOnce([
+      { big: long('9007199254740993'), max: long('9223372036854775807') },
+    ]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'c', operation: 'find' }),
+    );
+    expect(r.rows[0]).toEqual({
+      big: '9007199254740993',
+      max: '9223372036854775807',
+    });
+  });
+
+  it('serializes Binary subtype 4 (UUID) to canonical UUID string via toUUID()', async () => {
+    const uuidStr = '550e8400-e29b-41d4-a716-446655440000';
+    const bin = {
+      _bsontype: 'Binary',
+      sub_type: 4,
+      buffer: Buffer.from(uuidStr.replace(/-/g, ''), 'hex'),
+      toUUID: () => ({ toString: () => uuidStr }),
+    };
+    mockCursor.toArray.mockResolvedValueOnce([{ id: bin }]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'c', operation: 'find' }),
+    );
+    expect(r.rows[0]).toEqual({ id: uuidStr });
+  });
+
+  it('serializes generic Binary as Binary(subType,base64) and handles empty buffer', async () => {
+    const bin = (sub: number, buf: Buffer) => ({
+      _bsontype: 'Binary',
+      sub_type: sub,
+      buffer: buf,
+    });
+    mockCursor.toArray.mockResolvedValueOnce([
+      { hello: bin(0, Buffer.from('hello')), empty: bin(0, Buffer.alloc(0)) },
+    ]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'c', operation: 'find' }),
+    );
+    expect(r.rows[0]).toEqual({
+      hello: 'Binary(0,aGVsbG8=)',
+      empty: 'Binary(0,)',
+    });
+  });
+
+  it('serializes Timestamp as Timestamp(t, i)', async () => {
+    const ts = { _bsontype: 'Timestamp', t: 1234567890, i: 1 };
+    mockCursor.toArray.mockResolvedValueOnce([{ ts }]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'c', operation: 'find' }),
+    );
+    expect(r.rows[0]).toEqual({ ts: 'Timestamp(1234567890, 1)' });
+  });
+
+  it('serializes BSON scalars nested inside arrays and sub-documents', async () => {
+    const dec = { _bsontype: 'Decimal128', toString: () => '1.5' };
+    const long = { _bsontype: 'Long', toString: () => '42' };
+    mockCursor.toArray.mockResolvedValueOnce([
+      { items: [{ price: dec }], stats: { count: long } },
+    ]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'c', operation: 'find' }),
+    );
+    expect(r.rows[0]).toEqual({
+      items: [{ price: '1.5' }],
+      stats: { count: '42' },
+    });
+  });
+});
+
 describe('MongoDBDriver.query – findOne / aggregate / count', () => {
   it('findOne returns at most one row', async () => {
     mockCollection.findOne.mockResolvedValueOnce({ name: 'Bob' });
--- a/server/src/drivers/mongodb.test.ts
+++ b/server/src/drivers/mongodb.test.ts
@@ -274,15 +274,27 @@ describe('MongoDBDriver.query – find', () => {
 });
 
 describe('MongoDBDriver.query – BSON scalar serialization', () => {
-  it('serializes Decimal128 to its numeric string (positive, negative, zero)', async () => {
+  it('serializes Decimal128 to its numeric string (positive, negative, zero, negative zero)', async () => {
     const dec = (s: string) => ({ _bsontype: 'Decimal128', toString: () => s });
     mockCursor.toArray.mockResolvedValueOnce([
-      { a: dec('123.45'), b: dec('-987654321.0000001'), c: dec('0') },
+      {
+        a: dec('123.45'),
+        b: dec('-987654321.0000001'),
+        c: dec('0'),
+        // bson preserves Decimal128's IEEE-754 sign bit, so '-0' is a real
+        // value distinct from '0' (mongosh / Compass render it as '-0').
+        nz: dec('-0'),
+      },
     ]);
     const r = await makeDriver().query(
       JSON.stringify({ collection: 'c', operation: 'find' }),
     );
-    expect(r.rows[0]).toEqual({ a: '123.45', b: '-987654321.0000001', c: '0' });
+    expect(r.rows[0]).toEqual({
+      a: '123.45',
+      b: '-987654321.0000001',
+      c: '0',
+      nz: '-0',
+    });
   });
 
   it('serializes Long to its string form (preserves precision past Number.MAX_SAFE_INTEGER)', async () => {
@@ -328,7 +340,7 @@ describe('MongoDBDriver.query – BSON scalar serialization', () => {
     );
     expect(r.rows[0]).toEqual({
       hello: 'Binary(0,aGVsbG8=)',
-      empty: 'Binary(0,)',
+      empty: 'Binary(0,<empty>)',
     });
   });
 
--- a/server/src/drivers/mongodb.ts
+++ b/server/src/drivers/mongodb.ts
@@ -49,6 +49,29 @@ function buildMongoUri(config: ConnectionConfig): string {
   return `mongodb://${auth}${config.host}:${config.port}/`;
 }
 
+// Detect bson scalar types via their `_bsontype` tag. We avoid `instanceof`
+// here because users may bring their own `bson` build (e.g. a transitive
+// duplicate), which breaks prototype-chain checks; the `_bsontype` getter is
+// the project's documented identity marker.
+function bsonType(val: unknown): string | null {
+  if (val === null || typeof val !== 'object') return null;
+  const t = (val as { _bsontype?: unknown })._bsontype;
+  return typeof t === 'string' ? t : null;
+}
+
+function serializeBinary(val: {
+  sub_type: number;
+  buffer: Buffer | Uint8Array;
+  toUUID?: () => { toString(): string };
+}): string {
+  // Subtype 4 (and the legacy 3) are UUIDs; render canonical hyphenated form.
+  if (val.sub_type === 4 && typeof val.toUUID === 'function') {
+    return val.toUUID().toString();
+  }
+  const buf = Buffer.isBuffer(val.buffer) ? val.buffer : Buffer.from(val.buffer);
+  return `Binary(${val.sub_type},${buf.toString('base64')})`;
+}
+
 function serializeValue(val: unknown): unknown {
   if (val === null || val === undefined) return null;
   if (val instanceof ObjectId) return val.toHexString();
@@ -59,6 +82,20 @@ function serializeValue(val: unknown): unknown {
     return val.toISOString().replace('T', ' ').replace(/\.\d{3}Z$/, '');
   }
   if (Buffer.isBuffer(val)) return val.toString('hex');
+  const tag = bsonType(val);
+  if (tag) {
+    // Order matters: Timestamp extends Long, so check it first.
+    if (tag === 'Timestamp') {
+      const ts = val as { t: number; i: number };
+      return `Timestamp(${ts.t}, ${ts.i})`;
+    }
+    if (tag === 'Decimal128' || tag === 'Long') {
+      return (val as { toString(): string }).toString();
+    }
+    if (tag === 'Binary') {
+      return serializeBinary(val as Parameters<typeof serializeBinary>[0]);
+    }
+  }
   if (Array.isArray(val)) return val.map(serializeValue);
   if (typeof val === 'bigint') return val.toString();
   if (typeof val === 'object') {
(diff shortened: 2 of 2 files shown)


server/src/drivers/mongodb.ts
```diff
@@ -49,6 +49,29 @@ function buildMongoUri(config: ConnectionConfig): string {
...
+  const t = (val as { _bsontype?: unknown })._bsontype;
+  return typeof t === 'string' ? t : null;
+}
+
+function serializeBinary(val: {
+  sub_type: number;
+  buffer: Buffer | Uint8Array;
+  toUUID?: () => { toString(): string };
+}): string {
+  // Subtype 4 (and the legacy 3) are UUIDs; render canonical hyphenated form.
```
eugene-mobile: Comment says "Subtype 4 (and the legacy 3) are UUIDs" but the code only branches on `sub_type === 4`. Subtype 3 (legacy/BSON v0.x UUID, byte-swapped on .NET/Java drivers) falls through to the generic `Binary(3,<base64>)` path — which is arguably correct (its byte order is non-canonical so `toUUID()` would lie) but then the comment is wrong. Either drop the parenthetical or add an explicit subtype-3 branch (e.g. emit `LegacyUUID(<hex>)` so users can tell). As written, future-readers will assume subtype 3 is handled and be surprised.

server/src/drivers/mongodb.ts
```diff
@@ -49,6 +49,29 @@ function buildMongoUri(config: ConnectionConfig): string {
...
+  if (val.sub_type === 4 && typeof val.toUUID === 'function') {
+    return val.toUUID().toString();
+  }
+  const buf = Buffer.isBuffer(val.buffer) ? val.buffer : Buffer.from(val.buffer);
+  return `Binary(${val.sub_type},${buf.toString('base64')})`;
```
eugene-mobile: Non-blocking, but worth considering: there is no upper bound on Binary size. A document with a multi-MB GridFS chunk reference, an embedded image, or a subtype-6 ciphertext blob will base64-encode in full and ship over the wire to the grid, where it will then be word-wrapped into a single cell. That can easily be a 10x JSON bloat plus a frozen browser. Consider a cap (e.g. truncate buffers > 1KB and append `…(<N> bytes total)`), gated behind a constant so it's easy to tune. At minimum, file a follow-up.

Also: subtypes 6 (client-side field-level encryption) and 7 (sensitive) get rendered as plain `Binary(6,<base64>)`. Technically fine, but a user staring at ciphertext in the grid won't necessarily realize what they're looking at. A label like `Encrypted(<N> bytes)` for 6/7 might be friendlier — also a follow-up.

server/src/drivers/mongodb.ts
```diff
@@ -59,6 +82,20 @@ function serializeValue(val: unknown): unknown {
     return val.toISOString().replace('T', ' ').replace(/\.\d{3}Z$/, '');
   }
   if (Buffer.isBuffer(val)) return val.toString('hex');
+  const tag = bsonType(val);
+  if (tag) {
+    // Order matters: Timestamp extends Long, so check it first.
```
eugene-mobile: The `// Order matters` comment is great, but a refactor that converts this chain into a `switch (tag)` or sorts branches alphabetically would silently re-introduce the bug (Long would shadow Timestamp). Consider strengthening to `// MUST be checked before 'Long' — Timestamp extends Long in bson 7.x` and/or restructuring as a single switch where ordering is irrelevant by checking `tag === 'Timestamp'` first explicitly. Low priority but cheap insurance.

server/src/drivers/mongodb.ts
```diff
@@ -59,6 +82,20 @@ function serializeValue(val: unknown): unknown {
...
+    if (tag === 'Timestamp') {
+      const ts = val as { t: number; i: number };
+      return `Timestamp(${ts.t}, ${ts.i})`;
+    }
+    if (tag === 'Decimal128' || tag === 'Long') {
+      return (val as { toString(): string }).toString();
```
eugene-mobile, line 102: Duck-typing on `_bsontype` is the right call (avoids the dup-bson hazard you cite), but it does mean a legitimately-stored MongoDB document containing the literal field shape `{ _bsontype: "Long", toString: ... }` would be misinterpreted. Realistically this requires both a string field named `_bsontype` AND a `toString` override on a sub-document, which Mongo can't store as a function — so the practical attack surface is near-zero. Worth a one-line comment acknowledging this is best-effort, or a defensive `typeof val.toString === 'function' && val.toString !== Object.prototype.toString` guard before calling it. Non-blocking.

server/src/drivers/mongodb.test.ts
```diff
@@ -273,6 +273,90 @@ describe('MongoDBDriver.query – find', () => {
...
+      { hello: bin(0, Buffer.from('hello')), empty: bin(0, Buffer.alloc(0)) },
+    ]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'c', operation: 'find' }),
+    );
+    expect(r.rows[0]).toEqual({
+      hello: 'Binary(0,aGVsbG8=)',
+      empty: 'Binary(0,)',
+    });
+  });
```
eugene-mobile, line 345: `Binary(0,)` with the trailing comma reads like a malformed function call (and your snapshot literally locks that in). Consider a less ambiguous empty marker — e.g. `Binary(0,<empty>)` or `Binary(0,"")`. Whatever you choose, please update the rationale in the PR body so reviewers don't have to guess at the format contract.

server/src/drivers/mongodb.test.ts
```diff
@@ -273,6 +273,90 @@ describe('MongoDBDriver.query – find', () => {
...
+
+  it('serializes Timestamp as Timestamp(t, i)', async () => {
+    const ts = { _bsontype: 'Timestamp', t: 1234567890, i: 1 };
+    mockCursor.toArray.mockResolvedValueOnce([{ ts }]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'c', operation: 'find' }),
+    );
```
eugene-mobile, line 352: Test coverage gap: no assertions for `Decimal128.toString()` edge cases (`NaN`, `Infinity`, `-0` — bson emits these as literal strings) or for `Long.MIN_VALUE` (negative). Also no test that a `Binary` whose `.buffer` is a raw `Uint8Array` (not a Node `Buffer`) gets handled by the `Buffer.from(val.buffer)` fallback in `serializeBinary` — your mock only ever passes a Buffer. Cheap to add and would catch a real-world path, since the mongodb driver does occasionally hand back `Uint8Array`-backed Binaries.

More broadly: every test mocks the BSON types with POJOs that you control. That's fine for branch coverage, but it can't catch API drift (e.g. `sub_type` becoming `subType` in a future bson version, `toUUID()` throwing on malformed bytes, etc.). Given that this code path feeds the user-visible ResultsGrid, please run the browser smoke item that's currently unchecked in the PR body before merging — it's the only thing that exercises the real bson objects end-to-end. Skipping it on a pure-data-transform argument is reasonable in isolation, but the unit tests here are structurally unable to verify the duck-typing actually matches what real bson emits.

eugene-mobile (commented):
Solid, well-scoped fix. Logic is correct and tests recurse properly. A handful of edge-case gaps worth addressing before merge — most importantly the misleading subtype-3 comment, the missing length cap on Binary, and the value of an actual browser smoke given that this is a user-visible rendering change.

server/src/drivers/mongodb.ts
eugene-mobile: Done in 4f558f4 — rewrote the comment to match what the code does (subtype 3 has non-canonical, driver-specific byte order, so calling toUUID() on it would silently lie; we intentionally fall through to the generic Binary(3,<base64>) path). Picked the simpler fix over a LegacyUUID branch since legacy subtype-3 UUIDs are extremely rare in greenfield databases and the generic Binary form is honest about what we're showing.

server/src/drivers/mongodb.ts
eugene-mobile: Filed as #131 (Binary length cap + FLE/sensitive subtype labels). Agreed both are real concerns; deferring to a follow-up to keep this PR focused on the rendering correctness fix.

server/src/drivers/mongodb.ts
eugene-mobile: Done in 4f558f4 — strengthened to flag that a switch refactor or alphabetical sort would silently re-introduce the bug, since Timestamp extends Long in bson 7.x. Kept the if-chain structure (a switch wouldn't change the ordering hazard, just rename it).

server/src/drivers/mongodb.ts
eugene-mobile: Rejecting this one. The attack surface requires (a) a string field literally named `_bsontype` AND (b) a callable `toString` override on the same sub-document — and BSON can't store functions, so (b) is unreachable from real-world Mongo data. We're not a security boundary against a user injecting hostile data into their own database; any value persisted is theirs by definition. Adding a `typeof toString === 'function'` guard would only defend against synthetic in-memory shapes that can't survive a round-trip through the driver. Not worth the noise.

server/src/drivers/mongodb.test.ts
eugene-mobile: Done in 4f558f4 — empty buffers now render as `Binary(0,<empty>)` and the test fixture matches. Picked `<empty>` over `""` to keep the format consistent with the existing angle-bracket convention used elsewhere in the codebase for placeholder/sentinel tokens.

server/src/drivers/mongodb.test.ts
eugene-mobile: Browser smoke run against a real `mongo:7` container (port 27018), seeded a doc with all 5 BSON types (Decimal128, Long, Binary subtype 0, UUID subtype 4, Timestamp) plus a nested sub-document and array containing more Decimal128/Long. Connected via the API, ran `find {}`, every cell rendered as expected: `123.45`, `9007199254740993`, `Binary(0,aGVsbG8=)`, canonical UUID string, `Timestamp(1234567890, 1)`, recursed `-0.50` and `1.23` through nested. **No real-bson API drift detected** — `sub_type`, `_bsontype`, and `toUUID()` all match the duck-typed contract.

One observation: `NumberLong('999')` (small enough to fit safe-int range) gets auto-promoted by the driver to a plain JS `number` before reaching our serializer, so it lands in the result as `999` not `"999"`. That's correct — the string-form preservation only matters past `MAX_SAFE_INTEGER` (which `bigCount: 9007199254740993` covers).

On the coverage gap: added the **Decimal128 negative-zero** case in 4f558f4 (real edge — IEEE-754 sign bit, distinct from `'0'`). Skipping `NaN`/`Infinity` and `Long.MIN_VALUE` — `bson`'s `.toString()` is the same call site for all of them, so we'd be testing `bson`, not our code. The `Uint8Array`-buffer fallback is one branch in `serializeBinary` that the smoke run didn't exercise (the driver handed back `Buffer`); willing to add a unit test for it if you'd like, but feels like duplicate coverage of a one-line `Buffer.from` call.

server/src/drivers/mongodb.ts
```diff
@@ -59,6 +89,22 @@ function serializeValue(val: unknown): unknown {
     return val.toISOString().replace('T', ' ').replace(/\.\d{3}Z$/, '');
   }
   if (Buffer.isBuffer(val)) return val.toString('hex');
+  const tag = bsonType(val);
+  if (tag) {
+    // MUST come before Long — Timestamp extends Long in bson 7.x; reordering
+    // (or sorting these branches alphabetically in a refactor) would silently
+    // match Timestamp here and lose the (t, i) tuple.
```
eugene-mobile, line 96: Strengthened comment is good — explicitly naming the alphabetical-sort refactor hazard is exactly what a future reader needs. Non-blocking.

server/src/drivers/mongodb.ts
```diff
@@ -49,6 +49,36 @@ function buildMongoUri(config: ConnectionConfig): string {
...
+
+function serializeBinary(val: {
+  sub_type: number;
+  buffer: Buffer | Uint8Array;
+  toUUID?: () => { toString(): string };
+}): string {
+  // Subtype 4 is the canonical UUID; render the hyphenated form. Subtype 3
+  // (legacy UUID) has non-canonical, driver-specific byte order, so calling
+  // toUUID() on it would silently produce a wrong-looking string — we leave
+  // legacy UUIDs in the generic Binary(3,<base64>) form.
```
eugene-mobile, line 70: Comment now matches the code accurately (subtype 3 falls through to the generic Binary branch). The greenfield-rarity argument for skipping a `LegacyUUID(<hex>)` rendering is reasonable; if a user ever does hit one, the base64 form is at least lossless.

server/src/drivers/mongodb.ts
```diff
@@ -49,6 +49,36 @@ function buildMongoUri(config: ConnectionConfig): string {
...
+  if (val.sub_type === 4 && typeof val.toUUID === 'function') {
+    return val.toUUID().toString();
+  }
+  const buf = Buffer.isBuffer(val.buffer) ? val.buffer : Buffer.from(val.buffer);
+  const b64 = buf.toString('base64');
+  // Use an explicit empty marker so an empty buffer doesn't render as the
+  // ambiguous `Binary(0,)` (which reads like a malformed function call).
+  const payload = b64.length === 0 ? '<empty>' : b64;
+  return `Binary(${val.sub_type},${payload})`;
```
eugene-mobile, line 79: `Binary(0,<empty>)` reads cleanly; the angle-bracket convention is consistent with how the rest of the codebase signals sentinels.

server/src/drivers/mongodb.ts
```diff
@@ -49,6 +49,36 @@ function buildMongoUri(config: ConnectionConfig): string {
...
 
+// Detect bson scalar types via their `_bsontype` tag. We avoid `instanceof`
+// here because users may bring their own `bson` build (e.g. a transitive
+// duplicate), which breaks prototype-chain checks; the `_bsontype` getter is
+// the project's documented identity marker.
+function bsonType(val: unknown): string | null {
+  if (val === null || typeof val !== 'object') return null;
+  const t = (val as { _bsontype?: unknown })._bsontype;
+  return typeof t === 'string' ? t : null;
+}
```
eugene-mobile, line 60: Re the `_bsontype` spoofing rebuttal: the "BSON can't store functions" argument is right that there's no code-execution path — every dispatch branch except `Long`/`Decimal128` either reads scalar fields or, in the `Binary` case, calls `Buffer.from(undefined)` which throws and bubbles up. So the worst case is a malformed sibling field crashing a `find` (availability, not integrity). Acceptable to leave as-is, but worth a defensive `typeof val.toUUID === 'function'` style check around the `Binary` branch in a follow-up if you want to harden against denial-of-service from poisoned documents. Non-blocking.

server/src/drivers/mongodb.test.ts
```diff
@@ -273,6 +273,102 @@ describe('MongoDBDriver.query – find', () => {
...
+    ]);
+    const r = await makeDriver().query(
+      JSON.stringify({ collection: 'c', operation: 'find' }),
+    );
+    expect(r.rows[0]).toEqual({
+      a: '123.45',
+      b: '-987654321.0000001',
+      c: '0',
+      nz: '-0',
+    });
```
eugene-mobile, line 297: Decimal128 negative-zero coverage is a nice add — pins behaviour that would otherwise silently change if we ever swapped to a numeric coercion. Skipping NaN/Infinity/Long.MIN_VALUE is defensible: those branches reduce to `bson.toString()` and the smoke run already validated the live driver's output for the ordinary range.

eugene-mobile (commented):
Approve. All six review threads resolved or appropriately deferred; the browser smoke against real mongo:7 caught the precision-preservation behaviour that pure POJO mocks could have masked. One small follow-up nit on the spoofing rebuttal — see inline.

PR #132, opened 2026-05-01
fix: drag-aware backdrop dismiss for connection modal

### LINKED ISSUE #130: BUG: Click outside of login window closes the window and leads to the app not connected to any DB
If you open the first screen of the app (login window) and click outside of it - you get into the app - which is not functional as it's not connected to any DB. 
This is even more confusing because the same result is achieve by an attempt to clear a value in any field by using a mouse - when your selection gesture ends out outside of login window - which happens very often.

Lets make sure the click outside login window doesn't result in the closing of the window.

eugene-mobile:
Closes #130

## Summary

The connection modal dismissed on backdrop click and on the Escape key. Both paths dropped the user into a non-functional disconnected app. The drag-select-to-clear-a-field gesture (mousedown inside an input, mouseup past the modal edge) hit the same backdrop click and dismissed the modal.

The fix preserves the deliberate **click-outside-to-close** UX while ignoring drags that merely *end* on the backdrop. Escape no longer dismisses.

## What changed

- Restored `onClick` on the overlay backdrop, but gated it with a drag-aware check: dismiss only when **all three** of `mousedown`, `mouseup`, and `click` land on the backdrop itself.
- Added `onMouseDown` / `onMouseUp` handlers on the overlay that record the press origin into a `useRef<boolean>`; the `onClick` reads and resets it.
- The `onMouseUp` handler clears the ref if the release lands inside the modal — covers the inverse case (mousedown on backdrop, mouseup inside the modal).
- Escape `useEffect` remains removed.
- The modal body no longer needs `onClick={(e) => e.stopPropagation()}` — the overlay's `e.target === e.currentTarget` guard already filters out bubbled clicks from inner elements.

## Behaviour matrix

| Gesture | Result |
|---|---|
| Click on backdrop (no drag) | Dismiss |
| Click inside the modal | Stay open |
| Drag from inside modal, release on backdrop | Stay open |
| Drag from backdrop, release inside modal | Stay open |
| Press Escape | Stay open |
| Cancel button | Dismiss |

## Test plan

- [x] Click on the dark backdrop — modal closes.
- [x] Drag from a text input past the modal edge and release on backdrop — modal stays open.
- [x] Press Escape with focus inside an input — modal stays open.
- [x] Drag from backdrop into the modal and release inside — modal stays open.
- [x] Click Cancel — modal closes.
- [x] `npm run build` clean.

### DIFF
--- a/src/components/ConnectionManager.tsx
+++ b/src/components/ConnectionManager.tsx
@@ -126,13 +126,6 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  useEffect(() => {
-    if (!onDismiss || isConnecting) return;
-    const key = (e: KeyboardEvent) => { if (e.key === 'Escape') onDismiss(); };
-    window.addEventListener('keydown', key);
-    return () => window.removeEventListener('keydown', key);
-  }, [onDismiss, isConnecting]);
-
   const isMongoUri = form.type === 'mongodb' && form.mongoMode === 'uri';
 
   const buildSubmitForm = (): ConnectionForm => {
@@ -247,8 +240,8 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
   });
 
   return (
-    <div style={s.overlay} onClick={!isConnecting ? onDismiss : undefined}>
-      <div style={s.modal} onClick={(e) => e.stopPropagation()}>
+    <div style={s.overlay}>
+      <div style={s.modal}>
         <div style={s.header}>
           <svg width="20" height="20" viewBox="0 0 40 40" fill="none">
             <path d="M6 6 C6 6, 20 2, 20 20 C20 38, 6 34, 6 34" stroke={t.accent} strokeWidth="2.5" strokeLinecap="round"/>
--- a/src/components/ConnectionManager.tsx
+++ b/src/components/ConnectionManager.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import type { CSSProperties } from 'react';
 import type { Theme } from '../theme';
 import { api } from '../api';
@@ -104,6 +104,11 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
   const [testing, setTesting] = useState(false);
   const [testResult, setTestResult] = useState<{ ok: true } | { ok: false; error: string } | null>(null);
   const [canSavePassword, setCanSavePassword] = useState(false);
+
+  // Track whether a click began on the backdrop. Only dismiss when both mousedown
+  // and click landed on the backdrop — a drag that starts inside the modal and
+  // releases on the backdrop (e.g. text-selection) must not close the modal.
+  const mouseDownOnBackdrop = useRef(false);
 
   useEffect(() => {
     let cancelled = false;
@@ -240,7 +245,17 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
   });
 
   return (
-    <div style={s.overlay}>
+    <div
+      style={s.overlay}
+      onMouseDown={(e) => { mouseDownOnBackdrop.current = e.target === e.currentTarget; }}
+      onMouseUp={(e) => { if (e.target !== e.currentTarget) mouseDownOnBackdrop.current = false; }}
+      onClick={(e) => {
+        const started = mouseDownOnBackdrop.current;
+        mouseDownOnBackdrop.current = false;
+        if (!onDismiss || isConnecting) return;
+        if (started && e.target === e.currentTarget) onDismiss();
+      }}
+    >
       <div style={s.modal}>
         <div style={s.header}>
           <svg width="20" height="20" viewBox="0 0 40 40" fill="none">


src/components/ConnectionManager.tsx
```diff
@@ -247,8 +245,18 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
   });
 
   return (
-    <div style={s.overlay} onClick={!isConnecting ? onDismiss : undefined}>
-      <div style={s.modal} onClick={(e) => e.stopPropagation()}>
+    <div
+      style={s.overlay}
+      onMouseDown={(e) => { mouseDownOnBackdrop.current = e.target === e.currentTarget; }}
+      onMouseUp={(e) => { if (e.target !== e.currentTarget) mouseDownOnBackdrop.current = false; }}
```
eugene-mobile, line 251: Non-blocking — input modality. `onMouseDown` / `onMouseUp` / `onClick` cover mouse and (via synthesized mouse events) most touch, but they miss pen / pointer-only paths on devices where the browser emits only pointer events. They also fire for right/middle clicks. The drag-aware overlay-dismiss pattern is most robustly written with `onPointerDown` / `onPointerUp` plus a `e.button === 0` guard. Worth a follow-up rather than blocking — primary surface is desktop and Cancel covers the touch path.

src/components/ConnectionManager.tsx
```diff
@@ -247,8 +245,18 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
...
+      onClick={(e) => {
+        const started = mouseDownOnBackdrop.current;
+        mouseDownOnBackdrop.current = false;
+        if (!onDismiss || isConnecting) return;
```
eugene-mobile, line 255: Minor asymmetry — `onMouseDown` / `onMouseUp` still mutate `mouseDownOnBackdrop.current` while `isConnecting` is true, and line 254 has already cleared it before the guard at 255 bails out. State stays consistent because the next press resets the ref, but the guard is `onClick`-only by design — worth a comment if you want the next reader to notice.

src/components/ConnectionManager.tsx
```diff
@@ -247,8 +245,18 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
...
+        if (started && e.target === e.currentTarget) onDismiss();
```
eugene-mobile, line 256: Accessibility regression worth calling out in the PR body. The overlay has no `role="dialog"` / `aria-modal="true"` / focus trap, and Escape-to-dismiss is now removed entirely — keyboard-only users have no way to dismiss the modal except by tabbing to Cancel. The user explicitly asked for Escape to be disabled (per the second commit message), so this is by design, but documenting the tradeoff in the PR body would prevent a future a11y pass from re-adding Escape and reintroducing #130. A middle ground: keep Escape, but only when no input has focus — most accidental Escapes happen while typing.

src/components/ConnectionManager.tsx
```diff
@@ -247,8 +245,18 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
...
+      }}
+    >
+      <div style={s.modal}>
```
eugene-mobile, line 259: Removing `onClick={(e) => e.stopPropagation()}` from the modal body is correct (the overlay's `e.target === e.currentTarget` guard does the work), and it means inner clicks now bubble to the overlay's `onMouseDown` / `onMouseUp` / `onClick` and reset `mouseDownOnBackdrop.current` — the desired outcome. Confirmed there's no ancestor listener in `App.tsx` that newly receives these bubbled clicks.

src/components/ConnectionManager.tsx
```diff
@@ -105,6 +105,11 @@ export function ConnectionManager({ onConnect, isConnecting, error, onDismiss, t
   const [testResult, setTestResult] = useState<{ ok: true } | { ok: false; error: string } | null>(null);
   const [canSavePassword, setCanSavePassword] = useState(false);
 
+  // Track whether a click began on the backdrop. Only dismiss when both mousedown
+  // and click landed on the backdrop — a drag that starts inside the modal and
+  // releases on the backdrop (e.g. text-selection) must not close the modal.
+  const mouseDownOnBackdrop = useRef(false);
```
eugene-mobile, line 111: Lifecycle is sound: the modal is conditionally mounted in `App.tsx` (`{!connected && showConnectionModal && <ConnectionManager .../>}`), so the ref re-initializes per open. Within a single open it's reset both by `onClick` (line 254) and by every fresh `onMouseDown` (line 250). No staleness risk.

eugene-mobile (commented):
Drag-aware backdrop dismiss is implemented correctly and the smoke matrix covers the cases that matter. A few non-blocking observations:

**Dead code from the rework** — `src/components/ConnectionManager.tsx:341` still calls `e.nativeEvent.stopImmediatePropagation()` on the autocomplete input's Escape handler. That was there to prevent the now-removed window-level Escape listener from also firing and dismissing the modal. With the window listener gone this is dead code; safe to leave, but a brief comment (or removal) would prevent future confusion. The autocomplete Escape behaviour itself is unaffected — it never relied on the global handler.

**Two-commit history** — commit 1 removes everything, commit 2 restores drag-aware click-outside. Only the second state ships. A squash before merge would give `git blame` a single coherent reason for the change; equally fine to leave for audit trail. Author's call.

**Ref lifecycle** — `mouseDownOnBackdrop` is sound. The modal is conditionally mounted in `App.tsx:452`, so the ref re-initializes on each open; within a single open, it's reset both by `onClick` and by every `onMouseDown`. No staleness risk.

**Autocomplete dropdown interference** — verified: the suggestion list lives inside `s.modal`, so its mouse events bubble to the overlay handlers and correctly mark `mouseDownOnBackdrop.current = false`. No interference.

PR #133, opened 2026-05-01
feat: MongoDBDriver union-of-keys field sampling (#118)

### LINKED ISSUE #118: MongoDB: union-of-keys schema sampling for getSchema/getTable

eugene-mobile:
Closes #118.

Follow-up from PR #117 review (comment 3164075737).

## Summary

`MongoDBDriver` now infers collection schema by sampling up to 50 of the most-recent documents and taking the union of their keys, replacing the old single-`findOne()` pass. Both `getSchema` and `getTable` go through a shared `inferColumnsFromSample` helper.

## Acceptance criteria

- [x] `getSchema` and `getTable` sample N documents (default 50, module-level `SAMPLE_SIZE`)
- [x] Field list is the union of keys across the sample
- [x] `ColumnInfo.dataType` reflects the dominant type; conflicts surfaced via `comment: "types: a, b, ..."`
- [x] `nullable` is `false` only when the field is present in every sampled doc (`_id` stays hard-coded)
- [x] Existing tests still pass; new tests cover union-of-keys, nullable derivation, type-conflict comment, deterministic sort, and tie-breaking

## Design choices

- **Sample size 50** — picked at module scope; balances schema coverage vs. per-collection latency in `getSchema` fanout.
- **Sort `{_id: -1}` + `limit(50)`** — deterministic and recent-biased; uses the always-indexed `_id`.
- **Type conflicts** — most-frequent type wins, ties broken by first-seen order. The `comment` field lists every observed type in first-seen order so the order is stable across re-runs.
- **`_id`** — always `pk: true`, `nullable: false`. Empty collections still emit the placeholder `_id` column.

## Test plan

- [x] `cd server && npm test` — 109 passed (55 in `mongodb.test.ts`)
- [x] Live smoke against a temp `mongo:7` container, seeded with `[{name:'A', age:25}, {name:'B', email:'b@x.com'}, {name:'C', age:'thirty'}]`. `getSchema` returned `_id`, `name`, `age`, `email` (the old single-doc path would have missed `email`); `age.comment` was `"types: string, int"`; `name.nullable` was `false`, `age.nullable` and `email.nullable` were `true`.

### DIFF
--- a/server/src/drivers/mongodb.test.ts
+++ b/server/src/drivers/mongodb.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
+import type { Document } from 'mongodb';
 
 const mockCursor = {
   project: vi.fn(),
@@ -556,25 +557,50 @@ describe('MongoDBDriver – getSchemas', () => {
   });
 });
 
+/**
+ * Build a stub collection whose `find().sort().limit().toArray()` chain resolves
+ * to the supplied docs. The returned object exposes spies on each link so tests
+ * can assert deterministic-sort + sample-size wiring.
+ */
+function makeSampleColl(docs: Document[], count: number) {
+  const toArray = vi.fn().mockResolvedValue(docs);
+  const limit = vi.fn(() => ({ toArray }));
+  const sort = vi.fn(() => ({ limit }));
+  const find = vi.fn(() => ({ sort }));
+  return {
+    find,
+    sort,
+    limit,
+    toArray,
+    estimatedDocumentCount: vi.fn().mockResolvedValue(count),
+  };
+}
+
 describe('MongoDBDriver – getSchema (field inference)', () => {
-  it('samples one document per collection and marks columns as inferred', async () => {
+  it('samples documents with deterministic sort and SAMPLE_SIZE limit', async () => {
     mockListCursor.toArray.mockResolvedValueOnce([
       { name: 'users', type: 'collection' },
       { name: 'orders_v', type: 'view' },
     ]);
 
-    const usersColl = {
-      findOne: vi.fn().mockResolvedValueOnce({
-        _id: new ObjectIdCtor('aaaaaaaaaaaaaaaaaaaaaaaa'),
-        name: 'Ann',
-        age: 30,
-        joined: new Date('2024-01-01T00:00:00.000Z'),
-      }),
-      estimatedDocumentCount: vi.fn().mockResolvedValueOnce(7),
-    };
+    const usersColl = makeSampleColl(
+      [
+        {
+          _id: new ObjectIdCtor('aaaaaaaaaaaaaaaaaaaaaaaa'),
+          name: 'Ann',
+          age: 30,
+          joined: new Date('2024-01-01T00:00:00.000Z'),
+        },
+      ],
+      7,
+    );
     mockDb.collection.mockReturnValueOnce(usersColl as unknown as typeof mockCollection);
 
     const schema = await makeDriver().getSchema('shop');
+    expect(usersColl.find).toHaveBeenCalledWith({});
+    expect(usersColl.sort).toHaveBeenCalledWith({ _id: -1 });
+    expect(usersColl.limit).toHaveBeenCalledWith(50);
+
     expect(schema.views).toEqual(['orders_v']);
     expect(schema.tables).toHaveLength(1);
     const t = schema.tables[0];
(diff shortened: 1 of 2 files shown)


eugene-mobile (commented):
Approve with one real correctness concern (null-valued fields skew the dominant-type vote) and a few smaller follow-up suggestions. Logic, tests, and the recent-bias / `_id` index design are sound; implementation matches the issue description.

server/src/drivers/mongodb.ts
```diff
@@ -412,17 +420,56 @@ export class MongoDBDriver implements DbDriver {
...
+          seen.add(name);
+          order.push(name);
+          typeCounts.set(name, new Map());
+          typeOrder.set(name, []);
+        }
+        presence.set(name, (presence.get(name) ?? 0) + 1);
+        const t = inferDataType(doc[name]);
+        const counts = typeCounts.get(name)!;
+        if (!counts.has(t)) typeOrder.get(name)!.push(t);
+        counts.set(t, (counts.get(t) ?? 0) + 1);
```
eugene-mobile, line 444: **Real correctness issue (non-blocking but worth a follow-up):** `inferDataType` returns the string `'null'` for `null` / `undefined` values (line 132), and that token is then counted alongside `int`, `string`, etc. when picking the dominant type.

Consequence: a field whose 50 sampled values are `[null, null, null, null, 1]` will report `dataType: 'null'` and `comment: 'types: null, int'`, even though the only non-null observed type is `int`. The user would also see `nullable: false` because the field is *present* in every doc — so the column ends up labelled "non-null with type null", which is contradictory.

Suggested fix: skip `null` from `typeCounts`/`typeOrder` (still count it for `presence`, but treat the field as nullable when *any* sampled value is null). Pseudocode:

```ts
presence.set(name, (presence.get(name) ?? 0) + 1);
const t = inferDataType(doc[name]);
if (t === 'null') {
  nullSeen.add(name);  // forces nullable: true
  continue;
}
```

Then `nullable = isId ? false : (presence < docs.length || nullSeen.has(name))`. A test with `[{x: null}, {x: 1}]` expecting `dataType: 'int'` and `nullable: true` would lock this in.

server/src/drivers/mongodb.ts
```diff
@@ -412,17 +420,56 @@ export class MongoDBDriver implements DbDriver {
...
+        firstSeen.length > 1 ? `types: ${firstSeen.join(', ')}` : '';
+      const isId = name === '_id';
       return {
         name,
         type: dataType,
         dataType,
-        pk: name === '_id',
-        nullable: name !== '_id',
+        pk: isId,
+        nullable: isId ? false : (presence.get(name)! < docs.length),
```
eugene-mobile, line 469: Worth a one-line code comment that `nullable` is *sample-derived*: a field present in all 50 sampled (most-recent) docs is reported `nullable: false` even if older docs in the collection don't have it. The comment block at line 41 explains the recent bias for *field discovery*; a parallel note here would make the asymmetry obvious to the next reader (and to anyone debugging a "why is this field non-nullable when I see nulls in production" report).

server/src/drivers/mongodb.ts
```diff
@@ -390,20 +394,24 @@ export class MongoDBDriver implements DbDriver {
...
       comment: '',
-      columns: this.inferColumns(sample),
+      columns: this.inferColumnsFromSample(docs),
     };
   }
 
-  private inferColumns(sample: Document | null): ColumnInfo[] {
-    if (!sample) {
+  private async sampleDocs(coll: ReturnType<Db['collection']>): Promise<Document[]> {
+    return coll.find({}).sort({ _id: -1 }).limit(SAMPLE_SIZE).toArray();
```
eugene-mobile, line 410: Minor: on a sharded cluster, `find({}).sort({_id: -1}).limit(50)` fans out to every shard and merges — usually fine, but `getSchema` runs this concurrently for every collection, so a database with N collections issues N concurrent shard-fanned sorts. For unsharded mongo this is a non-issue (single `_id` index scan). Probably acceptable for now; might be worth a TODO referencing `$sample` or `find().limit()` (natural order) as a fallback for large/sharded deployments.

server/src/drivers/mongodb.ts
```diff
@@ -38,6 +38,13 @@ interface MqlRequest {
 
 const SYSTEM_DBS = new Set(['admin', 'local', 'config']);
 
+// 50 docs is a compromise between schema coverage and per-collection latency
+// during getSchema (which fans out across every collection in the database).
+// We sort by `{_id: -1}` so the sample is deterministic and biased toward
+// recent writes — newer fields in evolving collections are more likely to
+// surface than if we picked an arbitrary natural-order page.
+const SAMPLE_SIZE = 50;
```
eugene-mobile, line 46: Consider making `SAMPLE_SIZE` overridable per driver instance (constructor arg or a `MONGODB_SAMPLE_SIZE` env var) for two reasons: (1) heavily-evolving collections benefit from larger samples, (2) tests / debugging benefit from shrinking it. Not blocking — can be a follow-up if/when a user reports the default is wrong.

server/src/drivers/mongodb.test.ts
```diff
@@ -584,21 +610,101 @@ describe('MongoDBDriver – getSchema (field inference)', () => {
...
+      2,
+    );
+    mockDb.collection.mockReturnValueOnce(usersColl as unknown as typeof mockCollection);
+
+    const schema = await makeDriver().getSchema('shop');
+    const v = schema.tables[0].columns.find((c) => c.name === 'v')!;
+    expect(v.dataType).toBe('string');
+    expect(v.comment).toBe('types: string, int');
+  });
+
```
eugene-mobile, line 700: Test coverage gaps worth filling in a follow-up (none are blocking):

1. **Sample cap**: no test verifies that supplying >50 docs to the cursor stub still produces a column set derived from only the first 50. (The current `limit(50)` assertion proves the *call* but not the *behavior*.)
2. **Null-valued field**: `[{x: null}, {x: 1}]` — would catch the dominant-type-is-`'null'` bug flagged in `mongodb.ts`.
3. **Date / ObjectId / nested-object `dataType`** in the union path — already covered for the single-doc case in the first test, but not in the union/conflict paths.
4. **Empty collection placeholder** is asserted but not the `comment === ''` invariant.

#1 and #2 would be the highest-value additions.

PR #137, opened 2026-05-02
fix: render <select> for enum columns in Insert Row dialog

eugene-mobile:
## Summary

- Enum columns in the Insert Row dialog were plain text inputs, requiring users to guess valid values
- Now detects `enum('a','b','c')` columns via `dataType === 'enum'` and renders a native `<select>` with the parsed options
- Columns with a default or nullable get a leading blank option (labelled `default: active` or `NULL`); required enums have no blank option

## Implementation

- Added `parseEnumValues(type: string): string[] | null` to `src/lib/sql.ts` — parses the full `enum('active','inactive','pending')` type string, handles escaped single quotes and backslashes
- 7 new unit tests for `parseEnumValues` (standard, single-value, escaped quotes/backslashes, non-enum returns null, case-insensitive)

## Test plan

- [ ] Open Insert Row on a table with an enum column — verify a `<select>` renders instead of a text input
- [ ] Confirm all enum values appear as options
- [ ] Nullable/defaulted enum: verify a blank leading option appears
- [ ] Required enum (NOT NULL, no default): verify no blank option
- [ ] Select a value, click Review SQL — verify the selected value appears correctly quoted in the INSERT statement
- [ ] `npm test` passes (19 tests)

### DIFF
--- a/src/components/InsertRowDialog.tsx
+++ b/src/components/InsertRowDialog.tsx
@@ -2,7 +2,7 @@ import { useState, useEffect, useMemo } from 'react';
 import type { CSSProperties } from 'react';
 import type { Theme } from '../theme';
 import type { SchemaColumn } from '../api';
-import { buildInsertSql } from '../lib/sql';
+import { buildInsertSql, parseEnumValues } from '../lib/sql';
 
 type CellValue = string | number | boolean | null;
 type EditKind = 'boolean' | 'number' | 'date' | 'datetime' | 'time' | 'text';
@@ -156,12 +156,14 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
               )}
               {editable.map(col => {
                 const kind = editKindForColumn(col);
+                const enumValues = col.dataType === 'enum' ? parseEnumValues(col.type) : null;
                 const inputType = kind === 'number' ? 'number'
                   : kind === 'date' ? 'date'
                   : kind === 'datetime' ? 'datetime-local'
                   : kind === 'time' ? 'time'
                   : 'text';
                 const required = !col.nullable && col.default === null;
+                const hasBlankOption = col.default !== null || col.nullable;
                 const blankLabel = col.default !== null ? `default: ${col.default}`
                   : col.nullable ? 'NULL'
                   : '';
@@ -181,11 +183,28 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
                           onChange={(e) => setDrafts(p => ({ ...p, [col.name]: e.target.value }))}
                           style={s.input}
                         >
-                          {(col.default !== null || col.nullable) && (
+                          {hasBlankOption && (
                             <option value="">{blankLabel}</option>
                           )}
                           <option value="true">true</option>
                           <option value="false">false</option>
+                        </select>
+                      ) : enumValues ? (
+                        <select
+                          value={drafts[col.name] ?? ''}
+                          onChange={(e) => setDrafts(p => ({ ...p, [col.name]: e.target.value }))}
+                          style={s.input}
+                        >
+                          {hasBlankOption ? (
+                            <option value="">{blankLabel}</option>
+                          ) : (
+                            // Disabled placeholder keeps the visible state aligned with the
+                            // empty draft, so the browser doesn't silently surface the first
+                            // enum value while still routing through the "Required column"
+                            // validator if the user hits Review SQL without picking.
+                            <option value="" disabled>— select —</option>
+                          )}
+                          {enumValues.map(v => <option key={v} value={v}>{v}</option>)}
                         </select>
                       ) : (
                         <input
--- a/src/lib/sql.test.ts
+++ b/src/lib/sql.test.ts
@@ -1,5 +1,36 @@
 import { describe, it, expect } from 'vitest';
-import { formatSqlValue, buildInsertSql } from './sql';
+import { formatSqlValue, buildInsertSql, parseEnumValues } from './sql';
+
+describe('parseEnumValues', () => {
+  it('parses a standard enum type', () => {
+    expect(parseEnumValues("enum('active','inactive','pending')")).toEqual(['active', 'inactive', 'pending']);
+  });
+
+  it('parses a single-value enum', () => {
+    expect(parseEnumValues("enum('only')")).toEqual(['only']);
+  });
+
+  it('unescapes single quotes inside values', () => {
+    expect(parseEnumValues("enum('O\\'Brien','Smith')")).toEqual(["O'Brien", 'Smith']);
+  });
+
+  it('unescapes backslashes inside values', () => {
+    expect(parseEnumValues("enum('C:\\\\path')")).toEqual(['C:\\path']);
+  });
+
+  it('returns null for non-enum types', () => {
+    expect(parseEnumValues('varchar(100)')).toBeNull();
+    expect(parseEnumValues('int')).toBeNull();
+  });
+
+  it('returns null for an empty string', () => {
+    expect(parseEnumValues('')).toBeNull();
+  });
+
+  it('is case-insensitive for the ENUM keyword', () => {
+    expect(parseEnumValues("ENUM('a','b')")).toEqual(['a', 'b']);
+  });
+});
 
 describe('formatSqlValue', () => {
   it('renders NULL for null', () => {
--- a/src/lib/sql.ts
+++ b/src/lib/sql.ts
@@ -1,5 +1,17 @@
 type CellValue = string | number | boolean | null;
 
+export function parseEnumValues(type: string): string[] | null {
+  const match = type.match(/^enum\((.+)\)$/i);
+  if (!match) return null;
+  const values: string[] = [];
+  const re = /'((?:[^'\\]|\\.)*)'/g;
+  let m;
+  while ((m = re.exec(match[1])) !== null) {
+    values.push(m[1].replace(/\\'/g, "'").replace(/\\\\/g, '\\'));
+  }
+  return values.length > 0 ? values : null;
+}
+
 export function formatSqlValue(v: CellValue): string {
   if (v === null) return 'NULL';
   if (typeof v === 'string') return `'${v.replace(/\\/g, '\\\\').replace(/'/g, "\\'")}'`;


src/components/InsertRowDialog.tsx
```diff
@@ -160,18 +162,33 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
...
+                          onChange={(e) => setDrafts(p => ({ ...p, [col.name]: e.target.value }))}
+                          style={s.input}
+                        >
+                          {hasBlankOption && (
+                            <option value=''>
+                              {col.default !== null ? `default: ${col.default}` : 'NULL'}
+                            </option>
+                          )}
+                          {enumValues.map(v => <option key={v} value={v}>{v}</option>)}
+                        </select>
```
eugene-mobile, line 208: **Bug — required enums always fail validation on first submit.**

`initialDrafts` seeds every column to `''` (line 43). For a required enum (`!nullable && default === null`), `hasBlankOption` is false, so no `<option value=''>` is rendered. The `<select value=''>` has no matching option, but the browser still visually selects the first option — without firing `onChange`. The user sees e.g. `active` selected, clicks **Review SQL**, and gets `Required column: status` even though a value appears chosen.

Fix: when an enum column is required, pre-seed `drafts[col.name]` to `enumValues[0]` in `initialDrafts` (or in a separate `useEffect` keyed off `editable`). Something like:
```ts
for (const c of editable) {
  const ev = parseEnumValues(c.type);
  const required = !c.nullable && c.default === null;
  out[c.name] = (ev && required) ? ev[0] : '';
}
```
This also matches what the user visually perceives, so behaviour is consistent.

src/components/InsertRowDialog.tsx
```diff
@@ -143,13 +143,15 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
                 </div>
               )}
               {editable.map(col => {
+                const enumValues = parseEnumValues(col.type);
```
eugene-mobile: Minor: the PR description says detection happens via `dataType === 'enum'`, but the implementation regex-parses `col.type` for every column on every render. Gating on `col.dataType === 'enum'` first would be cheaper and self-documenting:
```ts
const enumValues = col.dataType === 'enum' ? parseEnumValues(col.type) : null;
```
The server already lowercases `dataType` (see `server/src/drivers/mysql.ts:175`), so the comparison is safe. Non-blocking.

src/lib/sql.ts
```diff
@@ -1,5 +1,17 @@
 type CellValue = string | number | null;
 
+export function parseEnumValues(type: string): string[] | null {
+  const match = type.match(/^enum\((.+)\)$/i);
+  if (!match) return null;
+  const values: string[] = [];
+  const re = /'((?:[^'\\]|\\.)*)'/g;
+  let m;
+  while ((m = re.exec(match[1])) !== null) {
+    values.push(m[1].replace(/\\'/g, "'").replace(/\\\\/g, '\\'));
```
eugene-mobile, line 10: Non-blocking: this only handles `\'` escaping. MySQL also permits `''` (doubled single quote) as an escape inside string literals, and `INFORMATION_SCHEMA.COLUMNS.COLUMN_TYPE` can return either form depending on `sql_mode` (notably `NO_BACKSLASH_ESCAPES`). Consider adding a test for `enum('it''s')` and either supporting it or documenting the limitation. Also worth a follow-up: `SET('a','b','c')` columns share the same shape but aren't handled here — they'll fall through to a plain text input.

src/lib/sql.ts
```diff
@@ -1,5 +1,17 @@
 type CellValue = string | number | null;
 
+export function parseEnumValues(type: string): string[] | null {
+  const match = type.match(/^enum\((.+)\)$/i);
```
eugene-mobile, line 4: Minor: the regex `/^enum\((.+)\)$/i` is greedy and anchored, which is fine, but it will also match `enumeration(...)` if such a string ever appeared (it never will from MySQL, but the name suggests a stricter check). Combined with the `dataType === 'enum'` gate suggested above, this would be unambiguous.

eugene-mobile (commented):
Nice fix overall. One real UX bug worth fixing before merge (required-enum drafts initialise to '' but no option carries value=''), plus a few smaller observations.

eugene-mobile (commented):
Fix for the required-enum first-submit bug works, but the implementation introduces a UX regression (silent auto-selection masks the required-field cue) and duplicates logic across two call sites. A few smaller follow-ups noted inline.

src/components/InsertRowDialog.tsx
```diff
@@ -40,7 +40,15 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
...
     const out: Record<string, string> = {};
-    for (const c of editable) out[c.name] = '';
+    for (const c of editable) {
+      // Required enums render no blank-state option, so a draft of '' would
+      // visually show the first option selected while state still says blank
+      // — failing "Required column" validation on first submit. Pre-seed the
+      // first enum value so the visible state matches what we'll send.
+      const enumValues = parseEnumValues(c.type);
+      const required = !c.nullable && c.default === null;
+      out[c.name] = enumValues && required ? enumValues[0] : '';
```
eugene-mobile: **UX regression for required enums.** Pre-seeding `drafts[c.name] = enumValues[0]` fixes the validation bug, but it also silently chooses a value on the user's behalf. The `*` next to the column name now lies — there is no required-field cue at all because the field is already 'filled'. A user who clicks Review SQL without touching the dropdown will insert `enumValues[0]` without ever seeing it.

Consider one of:
1. Render a disabled placeholder option (`<option value='' disabled>-- Select --</option>`) and keep `drafts[c.name] = ''`. The existing `missing.push(c.name)` path in `buildPayload` then surfaces the required-column error correctly, matching how text inputs behave.
2. If you keep the pre-seed, at minimum drop the `*` for the seeded case so the UI is consistent with its own semantics.

Option 1 is more in line with how the rest of the dialog treats required fields (text inputs are blank and rely on validation).

src/components/InsertRowDialog.tsx
```diff
@@ -143,13 +151,15 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
...
               {editable.map(col => {
+                const enumValues = parseEnumValues(col.type);
                 const kind = editKindForDataType(col.dataType);
                 const inputType = kind === 'number' ? 'number'
                   : kind === 'date' ? 'date'
                   : kind === 'datetime' ? 'datetime-local'
                   : kind === 'time' ? 'time'
                   : 'text';
                 const required = !col.nullable && col.default === null;
+                const hasBlankOption = col.nullable || col.default !== null;
```
eugene-mobile: **Logic duplicated across two sites.** `required = !c.nullable && c.default === null` is computed both here (line 161) and in `initialDrafts` (line 49), and `hasBlankOption = col.nullable || col.default !== null` is the inverse of `required`. Two booleans expressing the same predicate is a drift hazard — if the rule for what counts as 'required' ever changes (e.g. for generated columns, virtual columns, or `ON UPDATE` defaults), one site will get updated and the other will silently disagree.

Extract a single helper, e.g. `function isRequired(c: SchemaColumn): boolean { return !c.nullable && c.default === null; }` in `lib/sql.ts` (next to `parseEnumValues`) or in this file, and use it in both places. Then `hasBlankOption` is just `!isRequired(col)`.

src/components/InsertRowDialog.tsx
```diff
@@ -143,13 +151,15 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
                 </div>
               )}
               {editable.map(col => {
+                const enumValues = parseEnumValues(col.type);
```
eugene-mobile: Non-blocking: `parseEnumValues(col.type)` now runs both inside `initialDrafts` (once per editable column on mount/columns-change) **and** here on every render of every column. Every keystroke in any field triggers `setDrafts`, which re-renders this map and re-parses the regex for every column. Cheap individually but O(columns) per keystroke.

A `useMemo` keyed on `editable` that maps `col.name -> enumValues | null` would give you a single parse pass and a single source of truth for both `initialDrafts` and the render. Combine with the duplicated-logic note below.

src/components/InsertRowDialog.tsx
```diff
@@ -160,18 +170,33 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
...
-                        step={kind === 'number' ? 'any' : undefined}
-                        onChange={(e) => setDrafts(p => ({ ...p, [col.name]: e.target.value }))}
-                        placeholder={
-                          col.default !== null ? `default: ${col.default}` :
-                          col.nullable ? 'NULL' :
-                          ''
-                        }
-                        style={s.input}
-                      />
+                      {enumValues ? (
```
eugene-mobile: Design: the per-column input branch (lines 173–199) is now a 27-line ternary nested 3 levels into JSX. Extracting a `<ColumnInput col={col} value={...} onChange={...} t={t} />` named helper would mirror the project guideline against large IIFE/ternary blocks in JSX, and would let you co-locate the enum/non-enum branching with the `editKindForDataType` switch.

Non-blocking — flagging because the ternary will only grow as more column types get specialized inputs (set, json, blob, etc.).

src/components/InsertRowDialog.tsx
```diff
@@ -160,18 +170,33 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
...
+                        <select
+                          value={drafts[col.name] ?? ''}
+                          onChange={(e) => setDrafts(p => ({ ...p, [col.name]: e.target.value }))}
+                          style={s.input}
+                        >
+                          {hasBlankOption && (
+                            <option value=''>
+                              {col.default !== null ? `default: ${col.default}` : 'NULL'}
```
eugene-mobile: Edge case: the blank option label uses `default: ${col.default}` verbatim. For enum columns, MySQL stores the default as a bare value (e.g. `active`, no quotes), so this renders `default: active` — fine. But if `col.default` ever contains literal angle brackets or quotes from a backend that doesn't strip them, it would render unescaped into the option text. React escapes text content automatically so it's not an XSS issue, just a cosmetic one.

More importantly: if `col.default` is not actually one of the enum values (rare but possible — e.g. someone altered the enum without updating the default), the user picking 'default: foo' from the blank option will skip the column in `buildPayload` and let MySQL emit a constraint error. That's arguably the right behavior, but it's not obvious from the UI. Worth a follow-up to validate `col.default` against `enumValues` and warn if mismatched.

src/lib/sql.test.ts
```diff
@@ -1,5 +1,36 @@
...
+  });
+
+  it('returns null for an empty string', () => {
+    expect(parseEnumValues('')).toBeNull();
+  });
+
+  it('is case-insensitive for the ENUM keyword', () => {
+    expect(parseEnumValues("ENUM('a','b')")).toEqual(['a', 'b']);
+  });
+});
```
eugene-mobile, line 33: Test gap: `parseEnumValues` is well-covered, but the new `InsertRowDialog` behavior — 'required enum gets seeded with `enumValues[0]`, nullable/defaulted enum gets seeded with `""`' — has zero coverage. Given this is exactly the case the PR exists to fix, a component test (or at minimum a unit test on an extracted `seedEnumDraft(col)` helper) would prevent regression. Without it, someone refactoring `initialDrafts` could re-introduce the original bug and CI would stay green.

PR #140, opened 2026-05-03
fix: render boolean columns as a dropdown in Insert Row dialog

eugene-mobile:
## Summary
- The Insert Row dialog rendered MySQL `tinyint(1)` columns as a number input, forcing users to type `0`/`1` even though the inline cell editor already uses a true/false dropdown for the same column type.
- Detect `tinyint(1)` and render a `<select>` with NULL/true/false (or true/false for NOT NULL), matching the inline editor.
- The blank sentinel `<option value=\"\">` preserves the existing default/null fallback semantics in `buildPayload` (blank → omit when there's a server default, or NULL when nullable).
- Widen `CellValue` and `formatSqlValue` (`src/lib/sql.ts`) to accept booleans so the generated SQL preview renders `true`/`false` literally.

## Test plan
- [ ] Insert into a table where `is_active TINYINT(1) NOT NULL DEFAULT 0` — dropdown shows `default: 0`, `true`, `false`; SQL preview renders the chosen literal.
- [ ] Insert into a table where `is_archived TINYINT(1) NULL` — dropdown shows `NULL`, `true`, `false`; leaving the default `NULL` produces `… is_archived) VALUES (…, NULL)`.
- [ ] Inline cell edit on the same boolean column still works as before.
- [ ] `npm test` passes (new `formatSqlValue` boolean cases included).
- [ ] `tsc -b` is clean.

### DIFF
--- a/src/components/InsertRowDialog.tsx
+++ b/src/components/InsertRowDialog.tsx
@@ -45,7 +45,13 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
   const editable = useMemo(() => columns.filter(c => !c.autoIncrement), [columns]);
   const initialDrafts = useMemo(() => {
     const out: Record<string, string> = {};
-    for (const c of editable) out[c.name] = '';
+    for (const c of editable) {
+      // A NOT NULL boolean with no server default has no blank-state option in
+      // its <select> (otherwise the user would see "— select —" pre-selected
+      // and submitting would always fail with "Required column"). Seed it to
+      // 'false' so the visible state matches what we'll send.
+      out[c.name] = isBooleanColumn(c) && !c.nullable && c.default === null ? 'false' : '';
+    }
     return out;
   }, [editable]);
 
@@ -175,7 +181,9 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
                           onChange={(e) => setDrafts(p => ({ ...p, [col.name]: e.target.value }))}
                           style={s.input}
                         >
-                          <option value="">{blankLabel || '— select —'}</option>
+                          {(col.default !== null || col.nullable) && (
+                            <option value="">{blankLabel}</option>
+                          )}
                           <option value="true">true</option>
                           <option value="false">false</option>
                         </select>
--- a/src/components/InsertRowDialog.tsx
+++ b/src/components/InsertRowDialog.tsx
@@ -4,11 +4,16 @@ import type { Theme } from '../theme';
 import type { SchemaColumn } from '../api';
 import { buildInsertSql } from '../lib/sql';
 
-type CellValue = string | number | null;
-type EditKind = 'number' | 'date' | 'datetime' | 'time' | 'text';
+type CellValue = string | number | boolean | null;
+type EditKind = 'boolean' | 'number' | 'date' | 'datetime' | 'time' | 'text';
 
-function editKindForDataType(dataType: string): EditKind {
-  switch (dataType) {
+function isBooleanColumn(c: SchemaColumn): boolean {
+  return c.type === 'tinyint(1)';
+}
+
+function editKindForColumn(c: SchemaColumn): EditKind {
+  if (isBooleanColumn(c)) return 'boolean';
+  switch (c.dataType) {
     case 'tinyint': case 'smallint': case 'mediumint': case 'int': case 'bigint':
     case 'decimal': case 'float': case 'double': case 'year':
       return 'number';
@@ -64,14 +69,15 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
     const missing: string[] = [];
     for (const c of editable) {
       const raw = drafts[c.name];
-      const kind = editKindForDataType(c.dataType);
+      const kind = editKindForColumn(c);
       if (raw === '' || raw === undefined) {
         if (c.default !== null) continue; // server-side default will fill in
         if (c.nullable) { values[c.name] = null; continue; }
         missing.push(c.name);
         continue;
       }
-      if (kind === 'number') values[c.name] = Number(raw);
+      if (kind === 'boolean') values[c.name] = raw === 'true';
+      else if (kind === 'number') values[c.name] = Number(raw);
       else if (kind === 'datetime') values[c.name] = fromDatetimeLocal(raw);
       else values[c.name] = raw;
     }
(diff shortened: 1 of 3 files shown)


src/components/InsertRowDialog.tsx
```diff
@@ -4,11 +4,16 @@ import type { Theme } from '../theme';
...
 
-type CellValue = string | number | null;
-type EditKind = 'number' | 'date' | 'datetime' | 'time' | 'text';
+type CellValue = string | number | boolean | null;
+type EditKind = 'boolean' | 'number' | 'date' | 'datetime' | 'time' | 'text';
 
-function editKindForDataType(dataType: string): EditKind {
-  switch (dataType) {
+function isBooleanColumn(c: SchemaColumn): boolean {
+  return c.type === 'tinyint(1)';
```
eugene-mobile, line 11: Non-blocking: this duplicates the `tinyint(1)` detection that `isBoolColumn` already does in `ResultsTable.tsx` (which checks `col?.type === 'tinyint(1)'` against `SchemaColumn`). Since both call sites are now using the exact same heuristic against the same `SchemaColumn.type` field, consider lifting a shared helper (e.g. `isBooleanColumn(c: SchemaColumn)` in `src/lib/sql.ts` or a new `src/lib/columnKind.ts`) so future tweaks (e.g. handling `tinyint(1) unsigned`, or BIT(1)) only need to change one place. Not required to merge.

Also worth a follow-up: the literal-string match will miss `tinyint(1) unsigned` if MySQL ever returns that for the column type. The current behavior matches the existing inline editor, so it isn't a regression — just flagging it.

src/components/InsertRowDialog.tsx
```diff
@@ -160,18 +169,26 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
...
-                        }
-                        style={s.input}
-                      />
+                      {kind === 'boolean' ? (
+                        <select
+                          value={drafts[col.name] ?? ''}
+                          onChange={(e) => setDrafts(p => ({ ...p, [col.name]: e.target.value }))}
+                          style={s.input}
+                        >
+                          <option value="">{blankLabel || '— select —'}</option>
```
eugene-mobile: Two consistency nits with the inline editor in `ResultsTable.tsx` (the boolean `<select>` around line 1484):

1. **Sentinel mismatch.** The inline editor uses `value="null"` for the NULL choice and decodes via `raw === 'null' ? null : raw === 'true'`. Here you use `value=""` and rely on the empty-string branch in `buildPayload`. Both work, but it means the two boolean editors in the app speak slightly different dialects. Pick one (probably the empty-string convention is fine, since the rest of `buildPayload` treats `''` as "no input"), and consider documenting it.

2. **NOT NULL + no default still shows a blank option as the default selection.** When `col.default === null && !col.nullable`, the dropdown opens on `<option value="">— select —</option>`, so a user who clicks Review SQL without touching the field hits the "Required column" error. The inline editor avoids this by omitting the NULL option entirely for non-nullable columns. Two options to match that behavior here:
   - Drop the blank option in the NOT NULL + no-default case and pre-seed `drafts[col.name]` with `'false'` (or `'true'`), so the field has a real value from render-time.
   - Or keep the blank option, but only render it when there is *something* meaningful to convey (`col.default !== null` or `col.nullable`).

Non-blocking, but worth fixing while we're here — it's the kind of papercut a user will hit on the very first NOT NULL boolean column.

src/components/InsertRowDialog.tsx
```diff
@@ -64,14 +69,15 @@ export function InsertRowDialog({ table, columns, onSubmit, onClose, t }: Insert
...
-      const kind = editKindForDataType(c.dataType);
+      const kind = editKindForColumn(c);
       if (raw === '' || raw === undefined) {
         if (c.default !== null) continue; // server-side default will fill in
         if (c.nullable) { values[c.name] = null; continue; }
         missing.push(c.name);
         continue;
       }
-      if (kind === 'number') values[c.name] = Number(raw);
+      if (kind === 'boolean') values[c.name] = raw === 'true';
```
eugene-mobile, line 85: Minor: `raw === 'true'` is correct for the values produced by the new `<select>` (`'true'` / `'false'`), but it'll silently coerce *any* other string to `false`. That can't actually happen given the only producer of `drafts[col.name]` for a boolean column is the select with three known options, so this is fine in practice — just be aware that adding a third option (e.g. `'unknown'`) later would silently become `false`. A defensive `raw === 'true' ? true : raw === 'false' ? false : null` would future-proof it, at the cost of one extra branch.

src/lib/sql.ts
```diff
@@ -1,4 +1,4 @@
-type CellValue = string | number | null;
+type CellValue = string | number | boolean | null;
 
 export function formatSqlValue(v: CellValue): string {
   if (v === null) return 'NULL';
```
eugene-mobile, line 4: Worth a one-line comment here (or in `formatSqlValue`) noting that booleans render as the unquoted SQL tokens `true` / `false`, which MySQL accepts as synonyms for `1` / `0`. That's load-bearing for the Insert Row preview's correctness on `tinyint(1)` columns and isn't obvious from the type signature alone — and it would prevent a future refactor from "helpfully" switching to `'1'`/`'0'` (which would break booleans serialized for any non-tinyint(1) BOOLEAN-typed column on a future Postgres backend, per `api.ts` already typing the `type` field with `'postgres'`).

src/lib/sql.test.ts
```diff
@@ -33,6 +33,11 @@ describe('formatSqlValue', () => {
   it('handles empty string', () => {
     expect(formatSqlValue('')).toBe("''");
   });
+
+  it('renders true/false for booleans', () => {
+    expect(formatSqlValue(true)).toBe('true');
+    expect(formatSqlValue(false)).toBe('false');
```
eugene-mobile, line 39: Consider adding a `buildInsertSql` test that exercises a boolean alongside the other types, e.g.:

```ts
expect(buildInsertSql('t', { is_active: true, name: 'x', age: null })).toBe(
  "INSERT INTO `t`\n  (`is_active`, `name`, `age`)\nVALUES\n  (true, 'x', NULL);"
);
```

The existing `formatSqlValue` cases prove the primitive serialization but not the integration with `buildInsertSql`'s comma-joining and column-list rendering for the new type.

eugene-mobile (commented):
Solid, small change. The approach (detect tinyint(1) → render select, widen CellValue/formatSqlValue to accept booleans) is consistent with the rest of the codebase and the test coverage matches the new surface. A few non-blocking notes about consistency with the inline editor in ResultsTable.tsx and a couple of edge cases around the leading blank option.

PR #141, opened 2026-05-03
feat: make schema browser sidebar resizable

eugene-mobile:
## Summary
- The schema browser sidebar had a fixed 224px width, so long table names were truncated with no way to see the full name. Add a drag-to-resize handle on the right edge.
- 5px hit target with `cursor: col-resize`, highlights in the accent color on hover and during drag.
- Width clamped between 160px (min) and 600px (max), default 224px.
- Persisted to \`localStorage\` (key: \`helix.schemaBrowser.width\`) on mouseup; restored on next load.

## Test plan
- [ ] Connect to any database with a long table name. Default 224px width truncates as before.
- [ ] Drag the right edge of the sidebar wider — long names become readable.
- [ ] Drag past 160px / 600px — width clamps at the limit.
- [ ] Reload the page — chosen width is restored.
- [ ] \`tsc -b\` is clean.

### DIFF
--- a/src/components/SchemaBrowser.tsx
+++ b/src/components/SchemaBrowser.tsx
@@ -93,24 +93,47 @@ export function SchemaBrowser({ schema, activeTable, onTableSelect, onSchemaChan
   const [resizing, setResizing] = useState(false);
   const dropInputRef = useRef<HTMLInputElement>(null);
 
+  // Holds the teardown for an in-flight drag so the unmount cleanup below can
+  // remove the window listeners and restore body styles even if the user
+  // disconnects (or otherwise unmounts the browser) mid-drag.
+  const dragCleanupRef = useRef<(() => void) | null>(null);
+  useEffect(() => () => { dragCleanupRef.current?.(); }, []);
+
   const onResizeStart = (e: React.MouseEvent) => {
+    if (e.button !== 0) return;
     e.preventDefault();
     setResizing(true);
     const startX = e.clientX;
     const startWidth = width;
     let lastWidth = startWidth;
+
+    // Lock document-level cursor + selection so the browser doesn't start
+    // selecting text — and flip the cursor away from col-resize — once the
+    // pointer leaves the 5px handle during a drag.
+    const prevUserSelect = document.body.style.userSelect;
+    const prevCursor = document.body.style.cursor;
+    document.body.style.userSelect = 'none';
+    document.body.style.cursor = 'col-resize';
+
     const onMove = (ev: MouseEvent) => {
       lastWidth = Math.max(MIN_WIDTH, Math.min(MAX_WIDTH, startWidth + (ev.clientX - startX)));
       setWidth(lastWidth);
     };
-    const onUp = () => {
-      setResizing(false);
+    const cleanup = () => {
       window.removeEventListener('mousemove', onMove);
       window.removeEventListener('mouseup', onUp);
+      document.body.style.userSelect = prevUserSelect;
+      document.body.style.cursor = prevCursor;
+      dragCleanupRef.current = null;
+    };
+    const onUp = () => {
+      setResizing(false);
+      cleanup();
       try { localStorage.setItem(WIDTH_KEY, String(lastWidth)); } catch { /* quota or disabled */ }
     };
     window.addEventListener('mousemove', onMove);
     window.addEventListener('mouseup', onUp);
+    dragCleanupRef.current = cleanup;
   };
 
   useEffect(() => {
--- a/src/components/SchemaBrowser.tsx
+++ b/src/components/SchemaBrowser.tsx
@@ -67,6 +67,21 @@ function NonTableGroupItems({ groupKey, label, schema, filter, t, emptyStyle, ro
   ))}</>;
 }
 
+const WIDTH_KEY = 'helix.schemaBrowser.width';
+const MIN_WIDTH = 160;
+const MAX_WIDTH = 600;
+const DEFAULT_WIDTH = 224;
+
+function readStoredWidth(): number {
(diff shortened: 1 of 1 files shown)


src/components/SchemaBrowser.tsx
```diff
@@ -67,15 +67,52 @@ function NonTableGroupItems({ groupKey, label, schema, filter, t, emptyStyle, ro
...
   const [filter, setFilter] = useState('');
   const [contextMenu, setContextMenu] = useState<{ x: number; y: number; name: string; type: ObjectType } | null>(null);
   const [showSchemaFor, setShowSchemaFor] = useState<{ name: string; type: ObjectType } | null>(null);
   const [confirmDrop, setConfirmDrop] = useState<ConfirmDrop | null>(null);
+  const [width, setWidth] = useState<number>(readStoredWidth);
+  const [resizing, setResizing] = useState(false);
   const dropInputRef = useRef<HTMLInputElement>(null);
 
+  const onResizeStart = (e: React.MouseEvent) => {
+    e.preventDefault();
```
eugene-mobile, line 104: Missing left-button guard. Right-clicking the handle (or middle-click) will start a resize drag and bind window listeners until the next mouseup. Add `if (e.button !== 0) return;` at the top of `onResizeStart`. While here, also note that contextmenu events on the handle aren't suppressed, so the browser context menu can interleave with a drag in progress.

src/components/SchemaBrowser.tsx
```diff
@@ -67,15 +67,52 @@ function NonTableGroupItems({ groupKey, label, schema, filter, t, emptyStyle, ro
...
+    setResizing(true);
+    const startX = e.clientX;
+    const startWidth = width;
+    let lastWidth = startWidth;
+    const onMove = (ev: MouseEvent) => {
+      lastWidth = Math.max(MIN_WIDTH, Math.min(MAX_WIDTH, startWidth + (ev.clientX - startX)));
+      setWidth(lastWidth);
+    };
+    const onUp = () => {
```
eugene-mobile: Drag selects text across the rest of the page. `userSelect: 'none'` is applied only to the 5px handle, but as soon as the cursor leaves it, mousemove fires over normal content and the browser starts selecting text (and the cursor flips away from `col-resize`). On resize start, set `document.body.style.userSelect = 'none'` and `document.body.style.cursor = 'col-resize'`, and restore them in `onUp`. This is a noticeable papercut once you drag fast.

src/components/SchemaBrowser.tsx
```diff
@@ -67,15 +67,52 @@ function NonTableGroupItems({ groupKey, label, schema, filter, t, emptyStyle, ro
...
+      setResizing(false);
+      window.removeEventListener('mousemove', onMove);
+      window.removeEventListener('mouseup', onUp);
+      try { localStorage.setItem(WIDTH_KEY, String(lastWidth)); } catch { /* quota or disabled */ }
+    };
+    window.addEventListener('mousemove', onMove);
```
eugene-mobile, line 134: Window listener leak on unmount mid-drag. If `SchemaBrowser` unmounts while the user is holding the mouse (e.g. switching connections from a context menu on another component, or a parent re-mount), the `mousemove`/`mouseup` listeners stay attached to `window` forever and `setResizing` runs on an unmounted component. Wrap the listener registration in a `useEffect` keyed on `resizing`, or store the cleanup in a ref and call it from a `useEffect` cleanup.

src/components/SchemaBrowser.tsx
```diff
@@ -100,7 +137,8 @@ export function SchemaBrowser({ schema, activeTable, onTableSelect, onSchemaChan
   const toggleTable = (name: string) => setExpandedTables(p => ({ ...p, [name]: !p[name] }));
 
   const s = {
-    root: { width: 224, background: t.bgSurface, borderRight: `1px solid ${t.border}`, display: 'flex', flexDirection: 'column', overflow: 'hidden', flexShrink: 0 } as CSSProperties,
+    root: { width, position: 'relative', background: t.bgSurface, borderRight: `1px solid ${t.border}`, display: 'flex', flexDirection: 'column', overflow: 'hidden', flexShrink: 0 } as CSSProperties,
```
eugene-mobile, line 163: Stored `MAX_WIDTH` of 600 isn't viewport-aware. On a small window (e.g. 800px wide) a previously persisted 600px sidebar leaves only 200px for the editor. Consider clamping against `window.innerWidth` (e.g. `Math.min(MAX_WIDTH, window.innerWidth - 320)`) on read and on resize, or listening to `window` resize. Non-blocking.

src/components/SchemaBrowser.tsx
```diff
@@ -131,6 +169,13 @@ export function SchemaBrowser({ schema, activeTable, onTableSelect, onSchemaChan
 
   return (
     <div style={s.root}>
+      <div
+        style={s.resizeHandle}
+        onMouseDown={onResizeStart}
+        onMouseEnter={(e) => { if (!resizing) e.currentTarget.style.background = t.accent; }}
```
eugene-mobile, line 198: Minor a11y / discoverability. The handle is a non-semantic `div`. Consider `role="separator"` with `aria-orientation="vertical"` and `aria-valuenow={width}`, plus keyboard support (arrow keys to nudge). Keyboard-only users currently can't resize. Non-blocking follow-up.

eugene-mobile (commented):
Nice, focused change. The resize logic itself works, but a few drag-UX issues are worth addressing before merge: page-wide text selection during drag, cursor flicker when the pointer leaves the 5px strip, no left-button guard, and a window listener leak if the component unmounts mid-drag.

PR #142, opened 2026-05-03
fix: scope query results per tab and dedupe table-click tabs

eugene-mobile:
## Summary
Two tab UX bugs:

1. **Query results disappeared when switching tabs.** \`results\`, \`isRunning\`, \`queryError\`, and \`execTime\` lived in App-level state, so switching tabs threw away whatever was on the previous tab. Move all four onto the \`Tab\` type and derive the displayed values from \`currentTab\`. Async handlers (run / insert / update / delete) capture the originating tab id at call time so a query that finishes after the user switches tabs still writes to the right tab.
2. **Clicking the same table twice opened duplicate tabs.** \`handleTableSelect\` now switches to the existing tab (matched by auto-generated name) instead of always calling \`addTab\`.

## Test plan
- [ ] Open table A in the sidebar, run \`SELECT *\`. Click table A again — same tab stays focused, no duplicate.
- [ ] Open table B in the sidebar, run \`SELECT *\`. Switch back to table A's tab — results from earlier are still there.
- [ ] Run a long query in tab A, switch to tab B before it finishes — results land in tab A when the query completes.
- [ ] Disconnect resets to a single empty \`query_1.sql\` tab.
- [ ] \`tsc -b\` is clean.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,6 +19,10 @@ interface Tab {
   name: string;
   query: string;
   modified?: boolean;
+  results?: QueryResults | null;
+  queryError?: string | null;
+  execTime?: number | null;
+  isRunning?: boolean;
 }
 
 const EMPTY_SCHEMA: SchemaData = { tables: [], views: [], procedures: [], triggers: [] };
@@ -65,13 +69,20 @@ export default function App() {
   ]);
   const [activeTab, setActiveTab] = useState('1');
   const tabCounter = useRef(1);
-  const [results, setResults] = useState<QueryResults | null>(null);
-  const [isRunning, setIsRunning] = useState(false);
-  const [queryError, setQueryError] = useState<string | null>(null);
-  const [execTime, setExecTime] = useState<number | null>(null);
   const [activeTable, setActiveTable] = useState<string | null>(null);
 
   const currentTab = tabs.find(tab => tab.id === activeTab)!;
+  const results = currentTab?.results ?? null;
+  const isRunning = currentTab?.isRunning ?? false;
+  const queryError = currentTab?.queryError ?? null;
+  const execTime = currentTab?.execTime ?? null;
+
+  // Per-tab state updater. Targets a specific tab id (captured at call time)
+  // so async query handlers write back to the originating tab even if the
+  // user has switched in the meantime.
+  const updateTab = useCallback((id: string, patch: Partial<Tab>) => {
+    setTabs(ts => ts.map(tb => tb.id === id ? { ...tb, ...patch } : tb));
+  }, []);
 
   const loadSchema = useCallback(async (schema: string) => {
     try {
@@ -153,17 +164,16 @@ export default function App() {
     setSchemas([]);
     setActiveSchema('');
     setSchemaData(EMPTY_SCHEMA);
-    setResults(null);
-    setQueryError(null);
-    setExecTime(null);
+    tabCounter.current = 1;
+    setTabs([{ id: '1', name: 'query_1.sql', query: '' }]);
+    setActiveTab('1');
     setActiveTable(null);
   };
 
   const handleDropTable = async (schema: string, table: string) => {
     await api.dropTable(schema, table);
     if (activeTable === table) setActiveTable(null);
-    setResults(null);
-    setQueryError(null);
+    updateTab(activeTab, { results: null, queryError: null });
     await loadSchema(schema);
   };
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -325,19 +339,22 @@ export default function App() {
 
   const handleTableSelect = (name: string) => {
     setActiveTable(name);
+    const tabName = queryMode === 'mql' ? `${name}.json` : `${name}.sql`;
+    const existing = tabs.find(tb => tb.name === tabName);
```
eugene-mobile: Deduping by tab `name` is fragile in two ways:

1. **Schema-blind**: `tabs.find(tb => tb.name === tabName)` doesn't consider `activeSchema`. If the user opens `users.sql` in schema `app_dev`, switches to schema `app_prod`, and clicks `users` again, they're silently sent back to the tab that was created in the original schema context. The query body (`SELECT * FROM \`users\` LIMIT 100`) still works because the table name is unqualified, but the tab no longer represents what the user clicked.
2. **User-renamed / collision-prone**: any tab a user (or another handler) names `users.sql` will be matched, regardless of whether it was created from this code path. There's no rename UI in this diff, but `handleReopenSavedQuery` already produces tabs named `${query.name}.sql` — a saved query named `users` would shadow the table click.

Consider attaching a structural marker to the tab (e.g. `sourceTable?: { schema: string; table: string }`) and matching on that instead of the display name. That also lets you regenerate the query body when `queryMode` changes, instead of producing a second `.json` tab for the same table on mode switch.

src/App.tsx
```diff
@@ -65,13 +69,20 @@ export default function App() {
   ]);
   const [activeTab, setActiveTab] = useState('1');
   const tabCounter = useRef(1);
-  const [results, setResults] = useState<QueryResults | null>(null);
-  const [isRunning, setIsRunning] = useState(false);
-  const [queryError, setQueryError] = useState<string | null>(null);
-  const [execTime, setExecTime] = useState<number | null>(null);
   const [activeTable, setActiveTable] = useState<string | null>(null);
 
   const currentTab = tabs.find(tab => tab.id === activeTab)!;
```
eugene-mobile, line 80: Nit: `tabs.find(...)!` here contradicts the defensive `currentTab?.` used on lines 75–78, 187, 231, and 432. The non-null assertion is reasonable in practice (close-tab logic immediately reassigns `activeTab`), but if it's truly invariant, drop the optional chaining everywhere; if the optional chaining is intentional, the assertion should be `?? undefined`. Mixing the two styles obscures the actual invariant.

src/App.tsx
```diff
@@ -259,7 +276,7 @@ export default function App() {
         if (entry) setHistory(prev => [entry, ...prev].slice(0, 100));
       }
     } finally {
-      setIsRunning(false);
+      updateTab(targetTab, { isRunning: false });
```
eugene-mobile, line 285: Three separate `updateTab` calls per run (start, success/error, finally) cause an intermediate render where `results` is already populated but `isRunning` is still `true`. ResultsTable may flash a spinner over completed results. You can fold the `isRunning: false` into the success/error patches and drop the `finally`, e.g.:

```ts
try {
  const res = ...;
  updateTab(targetTab, { results: ..., execTime: ..., isRunning: false });
} catch (err) {
  updateTab(targetTab, { queryError: message, isRunning: false });
}
```

Non-blocking, but it removes a render and makes the state machine easier to reason about.

eugene-mobile (commented):
Solid fix overall — moving per-tab transient state onto `Tab` and capturing `targetTab` at call time is the right shape. A few non-blocking concerns around the dedupe heuristic and a couple of minor consistency issues below.

src/App.tsx
```diff
@@ -325,19 +339,22 @@ export default function App() {
...
+      addTab(tabName, `SELECT *\nFROM \`${name}\`\nLIMIT 100;`);
     }
   };
 
   const switchTab = (id: string) => {
     setActiveTab(id);
-    setResults(null);
-    setQueryError(null);
-    setExecTime(null);
   };
```
eugene-mobile, line 373: Two related UX seams worth noting (neither blocks this PR):

1. `switchTab` no longer clears anything, but `activeTable` also isn't synced when switching to a tab that was created from a table click. Selecting `users.sql` from the tab strip leaves the SchemaBrowser highlighting whatever was last clicked from the sidebar, which can disagree with the visible query.
2. `switchTab` is now just `setActiveTab` — could be inlined, or kept as the seam for the sync logic in (1).

If you add a `sourceTable` field per the dedupe comment above, both can be solved together: on tab switch, set `activeTable` to `tab.sourceTable?.table ?? null`.

src/App.tsx
```diff
@@ -189,58 +202,62 @@ export default function App() {
...
+        updateTab(targetTab, { queryError: 'MQL request must be a JSON object (e.g. { "collection": "...", "operation": "find" }).' });
         return;
       }
     }
 
-    setIsRunning(true);
-    setResults(null);
-    setQueryError(null);
-    setExecTime(null);
+    updateTab(targetTab, { isRunning: true, results: null, queryError: null, execTime: null });
```
eugene-mobile, line 256: Minor: in the MQL parse-error early return (lines 241/245) you set `queryError` but leave any prior `results`/`execTime` from this tab visible. That matches pre-PR behavior so it's not a regression, but now that errors are tab-scoped it would be cheap to clear them at the same time:

```ts
updateTab(targetTab, { queryError: ..., results: null, execTime: null });
```

Otherwise a successful run followed by an MQL parse error shows stale rows next to the new error message.

PR #144, opened 2026-05-03
feat: run multiple SQL statements in one query

### LINKED ISSUE #143: Support running multiple SQL statements in one editor execution
### Problem

Submitting more than one statement in the query editor — e.g.

```sql
SELECT 1 AS a;
SELECT 2 AS b;
```

— fails with the raw MySQL syntax error:

> You have an error in your SQL syntax; ... near 'SELECT 2 AS b;' at line 1

The MySQL pool is created without `multipleStatements: true`, so the driver rejects anything past the first `;`. Postgres accepts multi-statement input over the simple query protocol but the route only surfaces one result set.

This is a common need — running a setup snippet, executing a small batch of DDL, or running several `SELECT`s side-by-side without re-clicking Run.

### Expected behaviour

- Editor accepts one or more semicolon-separated statements.
- Each statement runs against the active connection in order.
- One result tab per statement that returns rows; affectedRows / insertId for write statements; per-statement error if one fails.
- Total execution time still shown in the status bar.

### Scope

- MySQL driver: enable `multipleStatements` and split the multi-result response from `mysql2`.
- Postgres driver: surface every result set returned by the simple query protocol.
- `/api/query`: return an array of result sets.
- Frontend: render dynamic `Result 1 … Result N` tabs in the results panel.
- MongoDB: out of scope (mql mode is single-request).

### Out of scope

- "Run statement at cursor" — separate UX, can be a follow-up.
- Streaming partial results.

eugene-mobile:
Closes #143.

## Summary
- MySQL pool now opts in to `multipleStatements`; `MysqlDriver.queryAll` returns one `QueryResult` per statement (mysql2's parallel-arrays response).
- `PostgresDriver.queryAll` runs the SQL through the simple query protocol and fans the resulting `Array<pg.QueryResult>` out to one entry per statement; legacy `query()` collapses to the last result for the existing single-statement callers (insertRow, updateCell, deleteRow, …).
- `/api/query` calls `queryAll` for sql-mode drivers and returns a `results: QueryResult[]` array. The pre-existing `columns` / `rows` / `affectedRows` / `insertId` envelope is preserved from the first statement so old clients keep working unchanged.
- Frontend stores all result sets per tab and renders dynamic `Result 1 … Result N` tabs in the results panel; clicking switches the rendered grid (and the target of CRUD actions). Single-statement queries still render exactly one tab — no UX change.

## Test plan
- [x] `npm test` (frontend) — 20/20 pass
- [x] `cd server && npm test` — 113/113 pass, including a new route test for the `queryAll` path and three new MysqlDriver tests covering single-result, multi-result, and error paths
- [x] Manual: connected to local MySQL test DB and ran `SELECT 1 AS a; SELECT * FROM bool_test LIMIT 2; SELECT 'three' AS msg;` — three Result tabs render and switching between them updates the grid
- [x] Manual: ran a single-statement `SELECT *` to confirm the regression case still shows just one Result tab

### DIFF
--- a/server/src/drivers/mysql.test.ts
+++ b/server/src/drivers/mysql.test.ts
@@ -141,3 +141,43 @@ describe('MysqlDriver.recyclePool', () => {
     expect(opts.keepAliveInitialDelay).toBe(10_000);
   });
 });
+
+describe('MysqlDriver.queryAll – multi-statement results', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockPool.getConnection.mockResolvedValue(mockConn);
+  });
+
+  it('returns one result per statement when mysql2 returns parallel arrays', async () => {
+    // mysql2's shape for multi-statement responses: arrays-of-arrays for both rows and fields.
+    mockConn.query.mockResolvedValueOnce([
+      [
+        [{ a: 1 }],
+        [{ b: 2 }],
+      ],
+      [
+        [{ name: 'a' }],
+        [{ name: 'b' }],
+      ],
+    ]);
+    const results = await makeDriver().queryAll('SELECT 1 AS a; SELECT 2 AS b');
+    expect(results).toHaveLength(2);
+    expect(results[0].rows).toEqual([{ a: 1 }]);
+    expect(results[1].rows).toEqual([{ b: 2 }]);
+    expect(mockConn.release).toHaveBeenCalledTimes(1);
+  });
+
+  it('wraps a single-statement response in a one-element array', async () => {
+    mockConn.query.mockResolvedValueOnce([[{ id: 1 }], [{ name: 'id' }]]);
+    const results = await makeDriver().queryAll('SELECT 1 AS id');
+    expect(results).toHaveLength(1);
+    expect(results[0].rows).toEqual([{ id: 1 }]);
+    expect(mockConn.release).toHaveBeenCalledTimes(1);
+  });
+
+  it('releases the connection when queryAll rejects', async () => {
+    mockConn.query.mockRejectedValueOnce(new Error('multi boom'));
+    await expect(makeDriver().queryAll('SELECT 1; SELECT 2')).rejects.toThrow('multi boom');
+    expect(mockConn.release).toHaveBeenCalledTimes(1);
+  });
+});
--- a/server/src/drivers/mysql.ts
+++ b/server/src/drivers/mysql.ts
@@ -15,12 +15,72 @@ function buildMysqlPoolOptions(config: ConnectionConfig): mysql.PoolOptions {
     waitForConnections: true,
     connectionLimit: 5,
     connectTimeout: 10_000,
+    multipleStatements: true,
     // OS-level TCP keepalive — without this, dead sockets after macOS sleep
     // are only surfaced by the kernel's default ~2-hour timer, which causes
     // the first post-resume query to hang. See #145.
     enableKeepAlive: true,
     keepAliveInitialDelay: 10_000,
   };
+}
+
+const MYSQL_NOT_NULL_FLAG = 1;
+const MYSQL_PRI_KEY_FLAG = 2;
+const MYSQL_UNIQUE_KEY_FLAG = 4;
+
+function buildMysqlResult(rows: RowDataPacket[] | ResultSetHeader, fields: FieldPacket[]): QueryResult {
+  if (!Array.isArray(rows)) {
+    const result = rows as ResultSetHeader;
+    return {
+      rows: [],
+      columnMeta: [],
+      affectedRows: result.affectedRows ?? 0,
+      insertId: result.insertId ?? null,
+    };
+  }
+
+  const columnMeta: ColumnMeta[] = fields.map(f => {
+    let flagsNum = 0;
+    if (typeof f.flags === 'number') {
+      flagsNum = f.flags;
+    } else if (Array.isArray(f.flags)) {
+      if ((f.flags as string[]).includes('NOT_NULL')) flagsNum |= MYSQL_NOT_NULL_FLAG;
+      if ((f.flags as string[]).includes('PRI_KEY')) flagsNum |= MYSQL_PRI_KEY_FLAG;
+      if ((f.flags as string[]).includes('UNIQUE_KEY')) flagsNum |= MYSQL_UNIQUE_KEY_FLAG;
+    }
+    return {
+      name: f.name,
+      orgName: f.orgName ?? f.name,
+      table: f.table ?? '',
+      orgTable: f.orgTable ?? '',
+      pk: (flagsNum & MYSQL_PRI_KEY_FLAG) === MYSQL_PRI_KEY_FLAG,
+      unique: (flagsNum & MYSQL_UNIQUE_KEY_FLAG) === MYSQL_UNIQUE_KEY_FLAG,
+      notNull: (flagsNum & MYSQL_NOT_NULL_FLAG) === MYSQL_NOT_NULL_FLAG,
+      mysqlType: f.columnType ?? f.type ?? 0,
+    };
+  });
+
+  const columns = fields.map(f => f.name);
+  const serializedRows = (rows as RowDataPacket[]).map(row => {
+    const out: Record<string, unknown> = {};
+    for (const col of columns) {
+      const val = row[col];
+      if (val === null || val === undefined) {
+        out[col] = null;
+      } else if (Buffer.isBuffer(val)) {
+        out[col] = val.toString('hex');
+      } else if (val instanceof Date) {
+        out[col] = val.toISOString().replace('T', ' ').replace(/\.\d{3}Z$/, '');
+      } else if (typeof val === 'bigint') {
+        out[col] = val.toString();
+      } else {
+        out[col] = val;
+      }
+    }
+    return out;
+  });
+
+  return { rows: serializedRows, columnMeta };
 }
 
 export class MysqlDriver implements DbDriver {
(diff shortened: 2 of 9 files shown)


server/src/drivers/postgres.ts
```diff
@@ -136,6 +140,27 @@ export class PostgresDriver implements DbDriver {
...
+    let searchPathSet = false;
+    try {
+      if (schema) {
+        await client.query(`SET search_path TO ${this.escapeIdent(schema)}`);
+        searchPathSet = true;
+      }
+      // Always go through the simple query protocol (no values) so multi-statement
+      // SQL fans out to one result per statement.
+      const raw = await client.query(sql);
+      const results = Array.isArray(raw) ? (raw as pg.QueryResult[]) : [raw];
```
eugene-mobile, line 154: **Likely incorrect — needs runtime verification.** `pg.Client.prototype.query(sqlString)` returns a single `pg.QueryResult` even for multi-statement simple-protocol SQL; node-pg accumulates the **last** command's `rows`/`fields` into the resolved Result and discards the prior ones (see node-postgres issues #1488, #2024). The `Array.isArray(raw)` branch is therefore probably never taken, and `queryAll` for Postgres silently returns only the last statement's rows.

To actually fan out into per-statement results you have to submit a `Query` instance and listen for `row`/`end` events per command, e.g. `client.query(new pg.Query({ text: sql, rowMode: 'array' }))` and aggregate, or use a library like `pg-cursor`. There is no Postgres `queryAll` test in this PR exercising a real multi-statement response, so the bug isn't caught.

At minimum: add an integration test that runs `SELECT 1 AS a; SELECT 2 AS b` against a real Postgres and asserts `results.length === 2`. If it's 1, this branch needs a rewrite using the Submittable `Query` API.

server/src/drivers/postgres.ts
```diff
@@ -85,47 +124,12 @@ export class PostgresDriver implements DbDriver {
...
-        }
-        return out;
-      });
-
-      return { rows: serializedRows, columnMeta };
+      // node-pg's simple query protocol returns an array of results when the SQL
+      // contains multiple statements. `query()` keeps the legacy single-result
+      // contract — `queryAll` is the multi-statement entry point — so collapse
+      // an unexpected array down to its last element.
+      const single = Array.isArray(result) ? (result as pg.QueryResult[])[result.length - 1] : result;
```
eugene-mobile, line 131: Inconsistent collapse direction with MySQL: `MysqlDriver.query` (mysql.ts:159–162) keeps the **first** result on multi-statement input, but here Postgres keeps the **last**. Internal callers (insertRow/updateCell/deleteRow) only ever pass single statements so today this never fires, but if anything else is later routed through `query()` with multi-statement SQL, the two drivers will behave differently. Pick one (last is the only one with the post-DDL `affectedRows`/`insertId` you're likely to want, but mysql does the opposite). Worth either aligning them or asserting/throwing on the unexpected array shape so the inconsistency can't silently bite us.

server/src/drivers/mysql.ts
```diff
@@ -15,6 +15,7 @@ function buildMysqlPoolOptions(config: ConnectionConfig): mysql.PoolOptions {
     waitForConnections: true,
     connectionLimit: 5,
     connectTimeout: 10_000,
+    multipleStatements: true,
```
eugene-mobile, line 18: Enabling `multipleStatements: true` materially widens the SQL-injection blast radius across the entire driver. Today the only caller that ever forwards untrusted text to the server is `query()`/`queryAll` from the `/api/query` route (which is by-design SQL execution), but any future code path in this driver that string-interpolates a user-controlled value into a `conn.query(...)` call now allows the attacker to chain statements after the first one — previously mysql2 would simply reject the second statement.

Worth (a) leaving an explicit comment on this option warning future maintainers, and (b) considering whether the multi-statement support could instead be scoped to a dedicated connection acquired only for `queryAll` (`mysql.createConnection({ ...opts, multipleStatements: true })`) so the rest of the pool stays single-statement. Not a blocker, but the security posture changed and the PR doesn't call it out.

eugene-mobile (commented):
Multi-statement plumbing on the MySQL side is solid, but the Postgres `queryAll` likely doesn't actually return one result per statement, the `query()` shims on the two drivers collapse multi-statement responses inconsistently (first vs last), and `multipleStatements: true` widens the SQL-injection blast radius for any future caller that interpolates user input. Frontend per-statement `affectedRows` is not surfaced anywhere. A few coverage gaps (Postgres `queryAll`, recyclePool+multipleStatements interaction, error mid-stream).

server/src/drivers/postgres.ts
```diff
@@ -136,6 +140,27 @@ export class PostgresDriver implements DbDriver {
...
+      // SQL fans out to one result per statement.
+      const raw = await client.query(sql);
+      const results = Array.isArray(raw) ? (raw as pg.QueryResult[]) : [raw];
+      return results.map(buildPgResult);
+    } finally {
+      if (searchPathSet) {
+        try { await client.query('SET search_path TO DEFAULT'); } catch { /* fall through to release */ }
+      }
+      client.release();
+    }
```
eugene-mobile, line 161: **No tests cover this method.** `mysql.test.ts` got three new `queryAll` tests (single, multi, error) but `postgres.test.ts` has zero. Given the simple-query-protocol concern above, this is the path most likely to be quietly broken — please add at least:
- single-statement returns one-element array,
- multi-statement returns N elements (this one will surface the pg shape question),
- rejection still releases the client and resets `search_path`.

The `searchPathSet` reset uses a `try/catch` that swallows the error — fine, but worth asserting in a test that on a failed reset the client is still released.

server/src/routes/query.ts
```diff
@@ -43,24 +43,32 @@ export const postQuery: RequestHandler = async (req, res) => {
...
+    const first = results[0] ?? { columns: [], columnMeta: [], rows: [], affectedRows: 0, insertId: null };
     res.json({
-      columns: result.columnMeta.map(c => c.name),
-      columnMeta: result.columnMeta,
-      rows: result.rows,
+      columns: first.columns,
+      columnMeta: first.columnMeta,
+      rows: first.rows,
+      affectedRows: first.affectedRows,
+      insertId: first.insertId,
```
eugene-mobile, line 70: Behavioral change to the legacy envelope: previously, on `columnMeta.length === 0` the response had `affectedRows`/`insertId`; on a SELECT it did **not** include those fields at all. The new code unconditionally sets them (`affectedRows: 0, insertId: null`) on every response. Any client that branches on `'affectedRows' in res` or `res.affectedRows !== undefined` to distinguish DML from SELECT will now see DML on every SELECT. The frontend doesn't appear to do this, but if there are any external consumers (Electron CLI, recorded fixtures, etc.) this is a quiet breaking change. Either (a) only set them when `first.columnMeta.length === 0`, or (b) call this out explicitly in the PR description.

server/src/routes/query.ts
```diff
@@ -43,24 +43,32 @@ export const postQuery: RequestHandler = async (req, res) => {
 
   try {
     const start = Date.now();
-    const result = await driver.query(queryText, [], schema);
+    // sql-mode drivers expose `queryAll` for multi-statement support; mql-mode
+    // payloads are always a single request and use `query`.
+    const resultList = driver.queryMode === 'sql' && driver.queryAll
+      ? await driver.queryAll(queryText, schema)
+      : [await driver.query(queryText, [], schema)];
```
eugene-mobile, line 50: Error mid-stream is not tested or documented. If `queryAll` runs `INSERT ...; SELECT broken_syntax;` mysql2 will reject the whole call after committing the first statement (no implicit transaction), and the route returns 400 with no per-statement context — the caller can't tell that statement 1 succeeded. Worth at minimum a route-level test asserting that an error from `queryAll` still returns 400 with the original message (the existing `query()` error test doesn't cover the new `queryAll` branch), and ideally surfacing the partial successes (`results: [...successes]` plus `error`) so users don't lose track of half-applied DDL/DML.

Also, no test exists for `resultList = []` (empty after `queryAll`). The `??` fallback handles it, but mysql2 would normally throw on empty SQL — confirm the route doesn't 200 with an empty envelope on, e.g., `;;`.

server/src/drivers/mysql.ts
```diff
@@ -91,63 +151,41 @@ export class MysqlDriver implements DbDriver {
...
+      ];
+      // Single statement: mysql2 returns the result directly. Multi-statement: it
+      // returns parallel arrays — one rowset and one fields array per statement.
+      const isMulti = Array.isArray(rawRows) && rawRows.length > 0 && Array.isArray((rawRows as unknown[])[0]);
+      if (!isMulti) {
+        return [buildMysqlResult(rawRows as RowDataPacket[] | ResultSetHeader, rawFields as FieldPacket[])];
+      }
+      const rowSets = rawRows as (RowDataPacket[] | ResultSetHeader)[];
+      const fieldSets = rawFields as FieldPacket[][];
+      return rowSets.map((r, i) => buildMysqlResult(r, fieldSets[i] ?? []));
```
eugene-mobile, line 188: When `rowSets` and `fieldSets` lengths disagree (mysql2 has historically had bugs around this for `OUT` parameters / stored procedure result-of-results) the `fieldSets[i] ?? []` fallback hides the desync silently and produces a result with `columnMeta: []`, which the route then mirrors into the legacy envelope as if the statement were DML. Probably fine for the common case, but a `console.warn` (or better, just trusting `rowSets.length` and warning when `fieldSets.length !== rowSets.length`) would make this less mysterious if it ever happens in production.

src/App.tsx
```diff
@@ -253,21 +258,30 @@ export default function App() {
...
+        : [{ columns: res.columns, columnMeta: res.columnMeta, rows: res.rows }];
       updateTab(targetTab, {
-        results: { columns: res.columns, columnMeta: res.columnMeta, rows: res.rows },
+        results: sets[0],
+        resultSets: sets,
+        activeResultIndex: 0,
         execTime: res.executionTime,
       });
       if (connectionHost) {
+        const totalRows = sets.reduce((n, s) => n + s.rows.length, 0);
```
eugene-mobile, line 281: `totalRows` sums `rows.length` across every result set, so for `INSERT INTO t VALUES (...); SELECT * FROM t;` the history entry shows the SELECT's count (correct) but for `SELECT a FROM t1; SELECT b FROM t2;` it shows `t1.count + t2.count` rather than either statement's count. That's defensible but the previous behavior was `res.rows.length` (the first result's count). Worth a one-line comment explaining the choice, or pivot to `sets[0].rows.length` to match the legacy single-statement semantics for history.

src/App.tsx
```diff
@@ -253,21 +258,30 @@ export default function App() {
...
+      // synthesised into a single-element array so the rest of the app stays uniform.
+      const sets: QueryResults[] = res.results
+        ? res.results.map(r => ({ columns: r.columns, columnMeta: r.columnMeta, rows: r.rows as QueryResults['rows'] }))
+        : [{ columns: res.columns, columnMeta: res.columnMeta, rows: res.rows }];
       updateTab(targetTab, {
-        results: { columns: res.columns, columnMeta: res.columnMeta, rows: res.rows },
+        results: sets[0],
+        resultSets: sets,
+        activeResultIndex: 0,
         execTime: res.executionTime,
```
eugene-mobile, line 278: Per-statement `affectedRows` is captured server-side (each `r.affectedRows` in the route's `results` array) but is never displayed anywhere on the frontend — the results panel only renders `rows`/`columnMeta`. For a multi-statement run like `UPDATE ...; UPDATE ...; SELECT ...;` the user sees a single Result tab and no indication that the two updates ran or how many rows they touched. Consider rendering `affectedRows` in each `Result N` tab header, or at least on the per-tab status bar, so multi-statement DML isn't invisible.

src/components/ResultsTable.tsx
```diff
@@ -622,12 +626,25 @@ export function ResultsTable({ results, isRunning, error, executionTime, activeS
...
+        {(resultSets && resultSets.length > 1
+          ? resultSets.map((_, i) => ({ label: `Result ${i + 1}`, index: i }))
+          : [{ label: 'Result 1', index: 0 }]
+        ).map(({ label, index }) => {
+          const active = index === activeResultIndex;
+          return (
+            <div
+              key={index}
+              style={active ? s.tabActive : s.tabInactive}
+              onClick={() => { if (!active && onSelectResultIndex) onSelectResultIndex(index); }}
```
eugene-mobile, line 638: Click handler ignores its lack of a key/role only partially — `role={onSelectResultIndex ? 'button' : undefined}` is set but there's no `onKeyDown` for Space/Enter, no `tabIndex`, and no `aria-selected`/`aria-pressed`. Keyboard users can't switch result tabs. A proper `<button>` (with the tab styling) would also avoid the role/keyboard-handler maintenance — these are visually tabs, semantically buttons-acting-as-tabs.

server/src/drivers/mysql.test.ts
```diff
@@ -141,3 +141,43 @@ describe('MysqlDriver.recyclePool', () => {
...
+    // mysql2's shape for multi-statement responses: arrays-of-arrays for both rows and fields.
+    mockConn.query.mockResolvedValueOnce([
+      [
+        [{ a: 1 }],
+        [{ b: 2 }],
+      ],
+      [
+        [{ name: 'a' }],
+        [{ name: 'b' }],
+      ],
```
eugene-mobile, line 161: The mocked multi-statement shape is plausible but not actually what mysql2 emits — real mysql2 returns `OkPacket`/`ResultSetHeader` instances for non-SELECT statements within a multi-statement response (e.g. for `UPDATE ...; SELECT ...;` you get `[ResultSetHeader, [rows...]]` for rows and `[undefined, [fields...]]` for fields). The current tests only exercise SELECT+SELECT. Please add a mixed DML+SELECT test asserting that statement 0's `affectedRows`/`insertId` survive and statement 1's `rows`/`columnMeta` are populated — that's the path most likely to break in production.

PR #146, opened 2026-05-04
fix: keep query path alive across macOS sleep/resume

### LINKED ISSUE #145: Queries hang after macOS sleep/resume in Electron
### Problem

After waking the machine from sleep, the Electron build can no longer run any query — the request just hangs (or eventually fails with \`ECONNRESET\` / \`Lost connection to MySQL server\`). Reconnecting from the connection modal restores it.

### Root cause

Idle TCP sockets in the connection pool don't survive sleep on macOS. The kernel keeps the socket marked open after resume, but the MySQL/Postgres server (and any NAT in between) has dropped the connection long ago. On the next query:

1. \`pool.getConnection()\` returns a stale socket from the pool.
2. \`conn.query(...)\` writes onto a dead socket and never gets a response — the route hangs because there's no app-level read timeout.

Compounding factors in the current driver code:
- mysql2 pool is created without \`keepAliveInitialDelay\`, so dead-socket detection falls back to macOS's default ~2-hour TCP keepalive interval.
- \`pg.Pool\` is created without \`keepAlive\` enabled.
- No pre-flight ping before queries; no \`powerMonitor\` wiring in Electron, so the bundled server has no idea the machine just resumed.

This hits the Electron build hardest because the app stays running across sleep cycles, whereas a dev shell typically gets restarted.

### Proposed fix

1. Enable TCP keepalive on both pools so the kernel surfaces dead sockets in seconds rather than hours:
   - mysql2: \`enableKeepAlive: true, keepAliveInitialDelay: 10_000\`
   - pg: \`keepAlive: true, keepAliveInitialDelayMillis: 10_000\`
2. On Electron \`powerMonitor.on('resume')\`, message the bundled server to recycle the active pool — close it and rebuild from the saved config — so the very first post-resume query opens a fresh socket. Connection state in the UI is preserved.

### Out of scope

- Per-query pre-flight ping (could be a follow-up if (1) + (2) prove insufficient).
- Auto-reconnect after explicit \`disconnect()\`.
- MongoDB driver (its internal heartbeats already cover this).

eugene-mobile:
Closes #145.

## Summary
- Enable TCP keepalive on the MySQL and Postgres pools (\`enableKeepAlive\` / \`keepAlive\`, 10 s initial delay) so dead sockets after sleep are detected in seconds instead of the OS default ~2 hours.
- New optional \`recyclePool()\` on \`DbDriver\` (implemented for MySQL and Postgres) that closes the active pool and rebuilds it from the saved config — the user-visible connection state is preserved.
- The bundled server listens for a \`host-resumed\` message on its utility-process \`parentPort\`. \`electron/main.cjs\` posts that message from \`powerMonitor.on('resume')\`, so the very first post-resume query opens a fresh socket. Standalone runs without a parent port silently no-op.

## Test plan
- [x] \`cd server && npm test\` — 115/115 pass, including new coverage for both \`recyclePool\` paths and the keepalive options
- [x] \`npm test\` (frontend) — 20/20 pass
- [ ] Manual: connect via Electron to a remote MySQL, sleep the laptop for ≥30 s, resume, run a query — expect it to succeed without a Reconnect detour

### DIFF
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -1,6 +1,6 @@
 'use strict';
 
-const { app, BrowserWindow, Tray, Menu, nativeImage, utilityProcess, dialog, ipcMain, safeStorage } = require('electron');
+const { app, BrowserWindow, Tray, Menu, nativeImage, utilityProcess, dialog, ipcMain, safeStorage, powerMonitor } = require('electron');
 const path = require('path');
 const http = require('http');
 const fs = require('fs');
@@ -205,6 +205,15 @@ app.whenReady().then(async () => {
   }
   createWindow();
   createTray();
+
+  // Sleep kills idle TCP sockets in the pool — the OS still reports them as
+  // open, so the first post-resume query hangs on a dead socket. Notify the
+  // bundled server so it drops and rebuilds its pool. See #145. Dev mode runs
+  // its own server outside this process, so the message has no recipient and
+  // is silently dropped.
+  powerMonitor.on('resume', () => {
+    serverProc?.postMessage({ type: 'host-resumed' });
+  });
 
   app.on('activate', () => {
     if (win) win.show();
--- a/server/src/drivers/interface.ts
+++ b/server/src/drivers/interface.ts
@@ -91,7 +91,8 @@ export interface DbDriver {
    * Drop and rebuild any underlying connection pool. Called after the host
    * machine resumes from sleep so the next query opens a fresh socket instead
    * of using a dead one the OS still thinks is open. Drivers that don't pool
-   * (or self-heal) may omit this.
+   * (or self-heal) may omit this — e.g. MongoDB's SDAM monitor heartbeats every
+   * `heartbeatFrequencyMS` and re-establishes dead servers on its own.
    */
   recyclePool?(): Promise<void>;
   end(): Promise<void>;
--- a/server/src/drivers/interface.ts
+++ b/server/src/drivers/interface.ts
@@ -87,5 +87,12 @@ export interface DbDriver {
   /** " LIMIT N" for dialects that support it in DML; "" for those that don't (e.g. Postgres). */
   rowLimitClause(n: number): string;
   ping(): Promise<void>;
+  /**
+   * Drop and rebuild any underlying connection pool. Called after the host
+   * machine resumes from sleep so the next query opens a fresh socket instead
+   * of using a dead one the OS still thinks is open. Drivers that don't pool
+   * (or self-heal) may omit this.
+   */
+  recyclePool?(): Promise<void>;
   end(): Promise<void>;
 }
--- a/server/src/drivers/mysql.test.ts
+++ b/server/src/drivers/mysql.test.ts
@@ -52,17 +52,32 @@ describe('MysqlDriver.recyclePool', () => {
   beforeEach(() => {
     vi.clearAllMocks();
     mockPool.getConnection.mockResolvedValue(mockConn);
+    // recyclePool fire-and-forgets `old.end()` via `.catch()`, so the mock has
+    // to return a real promise; bare vi.fn() returns undefined.
+    mockPool.end.mockResolvedValue(undefined);
   });
 
-  it('replaces the pool and ends the old one', async () => {
-    const mysqlMod = (await import('mysql2/promise')).default as unknown as { createPool: ReturnType<typeof vi.fn> };
+  async function getCreatePoolMock() {
+    const mod = (await import('mysql2/promise')).default as unknown as { createPool: ReturnType<typeof vi.fn> };
+    return mod.createPool;
+  }
+
+  // Hand each createPool call a distinct pool instance so a test can assert
+  // that operations after recycle target the *new* pool, not the old one.
+  async function stubPoolsPerCreate(...pools: typeof mockPool[]) {
+    const cp = await getCreatePoolMock();
+    let i = 0;
+    cp.mockImplementation(() => pools[Math.min(i++, pools.length - 1)]);
+  }
+
+  it('builds a fresh pool and ends the old one', async () => {
+    const cp = await getCreatePoolMock();
     const driver = makeDriver();
-    expect(mysqlMod.createPool).toHaveBeenCalledTimes(1);
+    expect(cp).toHaveBeenCalledTimes(1);
 
     mockPool.end.mockResolvedValueOnce(undefined);
     await driver.recyclePool();
-    // One createPool call for the constructor, one for the recycle.
-    expect(mysqlMod.createPool).toHaveBeenCalledTimes(2);
+    expect(cp).toHaveBeenCalledTimes(2);
     expect(mockPool.end).toHaveBeenCalledTimes(1);
   });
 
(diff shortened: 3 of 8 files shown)


server/src/drivers/mysql.ts
```diff
@@ -2,24 +2,47 @@ import mysql from 'mysql2/promise';
...
+
+  /**
+   * Drop the current pool and rebuild it from the saved config. Used after the
+   * machine resumes from sleep — every pre-existing socket is dead but the
+   * pool would happily hand it back to the next query, hanging the request.
+   */
+  async recyclePool(): Promise<void> {
+    const old = this.pool;
+    this.pool = mysql.createPool(buildMysqlPoolOptions(this.config));
+    try { await old.end(); } catch { /* old pool sockets are likely dead — swallow */ }
```
eugene-mobile: Non-blocking, but worth thinking through: there is no synchronization between `recyclePool()` and an in-flight `query()`. The sequence is:

1. A request calls `pool.getConnection()` on the *old* pool and is awaiting the query.
2. `host-resumed` fires; `recyclePool()` swaps `this.pool` to a new pool and calls `old.end()`.
3. mysql2's `pool.end()` waits for active connections to be released, but if the in-flight query is hung on a dead socket (the very situation we're fixing), `old.end()` itself will hang until the kernel surfaces the dead socket — which thanks to `enableKeepAlive` is now ~10 s + probe interval rather than 2 hours, so this is much better than before, but still not instant.

Two small things that would tighten this up:
- Don't `await old.end()` — fire it off with `.catch(() => {})` so a slow drain on dead sockets doesn't block the message handler (or anything that awaits `recyclePool()` transitively).
- Optionally, after building the new pool, do a quick `getConnection()`/`release()` (or `ping()`) so the *next* user query doesn't pay the TCP+auth cost. The user explicitly flagged this in the PR description and I think it's worth it — the whole point of the fix is making the post-resume path feel snappy.

If you do pre-warm, don't await it from `recyclePool()` either; just kick it off.

server/src/drivers/mysql.ts
```diff
@@ -2,24 +2,47 @@ import mysql from 'mysql2/promise';
...
+  }
+
+  /**
+   * Drop the current pool and rebuild it from the saved config. Used after the
+   * machine resumes from sleep — every pre-existing socket is dead but the
+   * pool would happily hand it back to the next query, hanging the request.
+   */
+  async recyclePool(): Promise<void> {
+    const old = this.pool;
+    this.pool = mysql.createPool(buildMysqlPoolOptions(this.config));
```
eugene-mobile, line 49: If two `host-resumed` events arrive in quick succession (e.g., laptop wakes briefly then sleeps again, or any bug in `powerMonitor` double-firing), two `recyclePool()` calls run in parallel, each builds a new pool, and the first new pool gets orphaned with no `end()` ever called on it — its sockets stay open until GC. A simple guard with an in-flight promise (`if (this.recycling) return this.recycling;`) would make this idempotent against bursts.

server/src/drivers/postgres.ts
```diff
@@ -1,23 +1,42 @@
...
+    this.pool = new pg.Pool(buildPgPoolConfig(config));
+  }
+
+  /**
+   * Drop the current pool and rebuild it from the saved config. See `MysqlDriver.recyclePool`.
+   */
+  async recyclePool(): Promise<void> {
+    const old = this.pool;
+    this.pool = new pg.Pool(buildPgPoolConfig(this.config));
+    try { await old.end(); } catch { /* old clients may already be dead — swallow */ }
```
eugene-mobile: Same two notes as the mysql side: don't `await old.end()` (so a slow drain on dead clients doesn't block), and consider pre-warming the new pool with a quick `pool.connect().then(c => c.release())` so the next user query hits a warm socket. Pg's `pool.end()` semantics are actually stricter than mysql2's — it waits for *all* checked-out clients to be released and rejects new `connect()` calls — so awaiting it on dead clients is the more painful of the two.

server/src/drivers/mysql.test.ts
```diff
@@ -47,3 +47,38 @@ describe('MysqlDriver.query – connection release', () => {
...
+    await expect(driver.recyclePool()).resolves.toBeUndefined();
+  });
+
+  it('configures TCP keepalive on the pool so post-resume sockets are detected', async () => {
+    const mysqlMod = (await import('mysql2/promise')).default as unknown as { createPool: ReturnType<typeof vi.fn> };
+    mysqlMod.createPool.mockClear();
+    makeDriver();
+    expect(mysqlMod.createPool).toHaveBeenCalledTimes(1);
+    const opts = mysqlMod.createPool.mock.calls[0][0];
+    expect(opts.enableKeepAlive).toBe(true);
```
eugene-mobile, line 140: Test coverage gap: nothing asserts that operations after `recyclePool()` actually use the new pool. The current test verifies `createPool` was called twice and `end` was called, but a regression that drops the `this.pool = mysql.createPool(...)` assignment would still pass. Suggest one extra assertion — call `driver.query(...)` after `recyclePool()` and assert `getConnection` was called on the new pool instance, not the old one. Same for the postgres test.

server/src/index.ts
```diff
@@ -56,3 +57,18 @@ app.listen(PORT, () => {
   console.log(`Helix server running at http://localhost:${PORT}`);
   console.log(`MCP endpoint:        http://localhost:${PORT}/mcp`);
 });
+
+// When forked as an Electron utilityProcess, listen for the host process's
+// power events. After resume the active pool's sockets are dead — recycle so
+// the next query opens fresh ones. Standalone runs have no parentPort, so this
+// is a no-op outside Electron.
+const parentPort = (process as unknown as { parentPort?: { on(ev: 'message', fn: (msg: unknown) => void): void } }).parentPort;
```
eugene-mobile, line 65: Minor cleanup suggestion for the `process.parentPort` cast. Electron's utility-process parent port is typed in `electron`'s `.d.ts` as `MessagePortMain` and is exposed on `NodeJS.Process` via module augmentation when `@electron/types` (or just `electron`) is in scope at type-check time. Two cleaner options:

1. Add a tiny `server/src/parent-port.d.ts` with `declare global { namespace NodeJS { interface Process { parentPort?: { on(ev: 'message', fn: (msg: unknown) => void): void } } } }` — then this becomes `if (process.parentPort) { process.parentPort.on(...) }` with no cast.
2. Or factor the cast into a one-line helper `function getParentPort(): ParentPort | undefined { ... }` so the route file isn't carrying a structural cast inline.

Not blocking — the duck-typed cast is correct — but the `as unknown as { ... }` does drift if the message protocol grows beyond `host-resumed`.

server/src/index.ts
```diff
@@ -56,3 +57,18 @@ app.listen(PORT, () => {
...
+if (parentPort) {
+  parentPort.on('message', (msg: unknown) => {
+    if (msg && typeof msg === 'object' && (msg as { type?: string }).type === 'host-resumed') {
```
eugene-mobile, line 68: If the message protocol gets a second member, the inline `(msg as { type?: string }).type === 'host-resumed'` pattern will start to ugly up. Consider extracting a small `isHostResumedMessage(msg: unknown): boolean` type guard now while there's exactly one — it'll force every new message type to be added in one place.

electron/main.cjs
```diff
@@ -206,6 +206,15 @@ app.whenReady().then(async () => {
   createWindow();
   createTray();
 
+  // Sleep kills idle TCP sockets in the pool — the OS still reports them as
+  // open, so the first post-resume query hangs on a dead socket. Notify the
+  // bundled server so it drops and rebuilds its pool. See #145. Dev mode runs
+  // its own server outside this process, so the message has no recipient and
+  // is silently dropped.
+  powerMonitor.on('resume', () => {
+    serverProc?.postMessage({ type: 'host-resumed' });
```
eugene-mobile, line 215: Worth pairing with `powerMonitor.on('suspend', ...)` to record a timestamp / log line, purely for diagnosing future tickets where the user reports "queries hung after sleep" — having a server-side log of exactly when sleep/resume fired makes #145-class issues much easier to triage. Not blocking.

server/src/drivers/interface.ts
```diff
@@ -87,5 +87,12 @@ export interface DbDriver {
   /** " LIMIT N" for dialects that support it in DML; "" for those that don't (e.g. Postgres). */
   rowLimitClause(n: number): string;
   ping(): Promise<void>;
+  /**
+   * Drop and rebuild any underlying connection pool. Called after the host
+   * machine resumes from sleep so the next query opens a fresh socket instead
+   * of using a dead one the OS still thinks is open. Drivers that don't pool
+   * (or self-heal) may omit this.
+   */
+  recyclePool?(): Promise<void>;
```
eugene-mobile, line 97: Making `recyclePool` optional on the interface is the right call — MongoDB's SDAM monitor does heartbeat the pool every `heartbeatFrequencyMS` (default 10 s) and will mark dead servers as `Unknown` and re-establish, so it genuinely doesn't need the explicit nudge. Worth a one-line comment here on the interface noting *why* it's optional ("e.g. MongoDB's SDAM heartbeats already cover this") so the next person adding a driver doesn't have to dig through the issue thread to figure out whether they need to implement it.

eugene-mobile (commented):
Solid fix overall — keepalive + a pool rebuild is the right shape. A few non-blocking notes around the recycle path (in-flight query handling, missing pre-warm, concurrent recycle) and a couple of test-coverage gaps. Nothing here blocks merge given the manual verification.

PR #152, opened 2026-05-08
feat: sort default table query by single auto-inc PK descending

### LINKED ISSUE #151: Default table query: sort by single auto-incrementing PK descending
## What
When the user clicks a table in the schema browser, the default query opens as:

\`\`\`sql
SELECT *
FROM \`t\`
LIMIT 100;
\`\`\`

For the common case (audit-trail tables with a single \`AUTO_INCREMENT\` PK), people want to see the most recent rows first, so they typically have to add \`ORDER BY id DESC\` themselves.

## Why scope it tightly
\`ORDER BY pk DESC\` is meaningless or confusing for tables with:
- UUID / string / hash primary keys
- Composite primary keys
- Non-incrementing identity columns

## Proposed behavior
Append \`ORDER BY \\\`<pk>\\\` DESC\` only when the table has **exactly one PK column** AND that column is **auto-incrementing** (\`autoIncrement: true\` from the schema info — already populated by both MySQL and Postgres drivers). Otherwise, leave the template as-is.

Examples:
- \`users (id INT AUTO_INCREMENT PK)\` → \`SELECT * FROM \\\`users\\\` ORDER BY \\\`id\\\` DESC LIMIT 100;\`
- \`sessions (token UUID PK)\` → \`SELECT * FROM \\\`sessions\\\` LIMIT 100;\` (unchanged)
- \`order_items (order_id, line_no PK)\` → \`SELECT * FROM \\\`order_items\\\` LIMIT 100;\` (unchanged, composite PK)

eugene-mobile:
Closes #151.

## Summary
When clicking a table in the schema browser, append \`ORDER BY \\\`<pk>\\\` DESC\` to the default query — but only for tables where the PK is a single auto-incrementing column. UUID, string, and composite PKs continue to get the unsorted \`SELECT * FROM t LIMIT 100;\` template, since descending order is meaningless on them.

## Behavior
| Table shape | Default query |
|---|---|
| \`users (id INT AUTO_INCREMENT PK)\` | \`SELECT * FROM \\\`users\\\` ORDER BY \\\`id\\\` DESC LIMIT 100;\` |
| \`sessions (token UUID PK)\` | \`SELECT * FROM \\\`sessions\\\` LIMIT 100;\` |
| \`order_items (order_id, line_no PK)\` | \`SELECT * FROM \\\`order_items\\\` LIMIT 100;\` |
| Postgres \`accounts (id BIGSERIAL PK)\` | \`SELECT * FROM \\\`accounts\\\` ORDER BY \\\`id\\\` DESC LIMIT 100;\` |

\`ColumnInfo.autoIncrement\` is already populated by both drivers (MySQL: \`extra LIKE '%auto_increment%'\`; Postgres: \`column_default LIKE 'nextval(%' OR is_identity = 'YES'\`), so no schema-loading changes are needed.

## Implementation
- New helper \`buildDefaultTableQuery(table, pkColumn | null)\` in \`src/lib/sql.ts\` — pure, easy to unit-test, mirrors the existing \`buildInsertSql\` style.
- \`handleTableSelect\` in \`App.tsx\` looks up the table's columns from \`schemaData\`, filters PK columns, and passes the column name only when there's exactly one and it's auto-incrementing.

## Test plan
- [x] Unit tests for \`buildDefaultTableQuery\` (unsorted fallback, sorted variant, identifier escaping).
- [x] \`npm test\` (23 frontend tests pass).
- [x] \`npm run build\` (typecheck + vite build clean).
- [ ] Manual: click a table with auto-inc PK, confirm \`ORDER BY\` is added; click a UUID-PK table, confirm it isn't.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,6 +13,7 @@ import { saveConnection } from './savedConnections';
 import { electronAPI } from './electronAPI';
 import { addHistoryEntry, listHistory, deleteHistoryEntry, clearHistory, type HistoryEntry } from './queryHistory';
 import { listSavedQueries, saveQuery, deleteSavedQuery, renameSavedQuery, type SavedQuery } from './savedQueries';
+import { buildDefaultTableQuery } from './lib/sql';
 
 interface Tab {
   id: string;
@@ -369,7 +370,14 @@ export default function App() {
       const body = JSON.stringify({ collection: name, operation: 'find', filter: {}, limit: 100 }, null, 2);
       addTab(`${name}.json`, body, source);
     } else {
-      addTab(`${name}.sql`, `SELECT *\nFROM \`${name}\`\nLIMIT 100;`, source);
+      // Sort by the PK descending only when it's a single auto-incrementing
+      // column, so the user lands on the most-recent rows for the common case.
+      // UUID/string/composite PKs get the unsorted template — descending order
+      // would be meaningless or surprising there.
+      const tableInfo = schemaData.tables.find(t => t.name === name);
+      const pkCols = tableInfo?.columns.filter(c => c.pk) ?? [];
+      const autoIncPk = pkCols.length === 1 && pkCols[0].autoIncrement ? pkCols[0].name : null;
+      addTab(`${name}.sql`, buildDefaultTableQuery(name, autoIncPk), source);
     }
   };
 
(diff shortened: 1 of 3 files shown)


eugene-mobile (commented):
Looks good. The single-auto-inc-PK gate correctly filters out composite and non-auto-inc keys, the helper is pure and unit-tested, and the optional chain in `handleTableSelect` safely degrades to the unsorted template if the table can't be found in `schemaData`. Backtick escaping matches the pre-existing template — any Postgres-quoting concern is pre-existing, not introduced here.

PR #154, opened 2026-05-09
feat: persist editor tabs per connection across app restarts

### LINKED ISSUE #153: Persist open editor tabs per connection across app restarts
## What
Tabs are currently in-memory only — closing and reopening the app drops the user back to a single empty editor tab, even when reconnecting to the same database. Re-finding the right tables and re-typing in-progress queries is friction every session.

## Proposed scope (tab-only, no result data)
Persist \`{ name, query, sourceTable }\` per tab plus the active tab id, scoped per connection (same key strategy as \`queryHistory\` / \`savedQueries\` — \`helix.tabs.<connectionName>\` in localStorage).

On reconnect to the same connection, restore tabs and active selection. New tab IDs continue from \`max(restoredId)\` so they can't collide.

## Explicitly out of scope
**Result data** is not persisted. Caching potentially-MB rowsets on disk introduces real costs:
- privacy: query results sit on disk between sessions
- size: localStorage's ~5MB origin quota gets eaten quickly
- staleness UX: needs banners and re-run affordances to be honest

Re-running a query is fast and unambiguous. Easy to add later if it turns out to be missed.

## Edge cases handled
- Empty connection identifier → no-op (matches \`queryHistory\` behavior).
- Malformed JSON / wrong shape → return null, fall back to default tab.
- Per-tab entries with wrong types → skip the bad entries, keep the good ones.
- Disconnect resets in-memory tabs to default but leaves persisted state intact (so reconnect restores them).

eugene-mobile:
Closes #153.

## Summary
Reconnecting to the same database now restores the user's previous editor tabs (name, query, sourceTable) and active selection. Tabs are scoped per connection in \`localStorage\` under \`helix.tabs.<connectionName>\`, mirroring the \`queryHistory\` / \`savedQueries\` pattern.

Result data is **not** persisted — caching MB-scale rowsets on disk has privacy, storage-quota, and stale-data UX costs that aren't worth it given how fast re-running is.

## Implementation
- \`src/tabPersistence.ts\` — \`loadTabs\` / \`saveTabs\` / \`clearTabs\` with defensive parsing (malformed JSON returns null; per-entry type-check skips bad tabs while keeping good ones).
- \`App.tsx\`:
  - On connect, after schema setup but before flipping \`connected: true\`, call \`loadTabs(res.connectionName)\` and seed \`tabs\` / \`activeTab\` / \`tabCounter\` from it. The counter seeds from \`max(restoredId)\` so newly-created tabs can't collide with restored ones.
  - A new \`useEffect\` on \`[connected, connectionHost, tabs, activeTab]\` writes the tab snapshot back. Guarded by \`connected && connectionHost\` so disconnect's reset-to-default doesn't wipe the saved set — reconnecting picks them back up.

## Test plan
- [x] Unit tests for \`tabPersistence\` (round-trip, scoping per connection, malformed JSON, partial-entry skipping, empty-connection guard, clear).
- [x] \`npm test\` (32 frontend tests pass).
- [x] \`npm run build\` (typecheck + vite build clean).
- [ ] Manual in Electron: open several tabs (mix of table-clicks and ad-hoc queries), restart, reconnect — confirm the tab list and active tab come back.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,6 +14,7 @@ import { electronAPI } from './electronAPI';
 import { addHistoryEntry, listHistory, deleteHistoryEntry, clearHistory, type HistoryEntry } from './queryHistory';
 import { listSavedQueries, saveQuery, deleteSavedQuery, renameSavedQuery, type SavedQuery } from './savedQueries';
 import { buildDefaultTableQuery } from './lib/sql';
+import { loadTabs, saveTabs } from './tabPersistence';
 
 interface Tab {
   id: string;
@@ -128,6 +129,24 @@ export default function App() {
       setSavedQueries(listSavedQueries(res.connectionName));
       setSchemas(list);
       setActiveSchema(initial);
+
+      // Restore the user's previous tabs for this connection. We deliberately
+      // don't persist result data (privacy + size); the user re-runs to get
+      // fresh rows. Seed tabCounter from the highest restored id so new tabs
+      // can't collide with restored ones.
+      const persistedTabs = loadTabs(res.connectionName);
+      if (persistedTabs) {
+        setTabs(persistedTabs.tabs);
+        const restoredActive = persistedTabs.tabs.find(t => t.id === persistedTabs.activeTabId)
+          ? persistedTabs.activeTabId
+          : persistedTabs.tabs[0].id;
+        setActiveTab(restoredActive);
+        tabCounter.current = persistedTabs.tabs.reduce(
+          (max, t) => Math.max(max, parseInt(t.id, 10) || 0),
+          0,
+        );
+      }
+
       setConnected(true);
       setShowConnectionModal(false);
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -403,6 +422,17 @@ export default function App() {
     if (connected && activeSchema) loadSchema(activeSchema);
   }, [activeSchema, connected, loadSchema]);
 
+  // Persist the editor tabs (name + query + sourceTable only) per connection.
+  // Skips when not connected so disconnect's reset-to-default doesn't wipe the
+  // saved set — reconnecting the same database picks them back up.
+  useEffect(() => {
+    if (!connected || !connectionHost) return;
+    saveTabs(connectionHost, {
+      tabs: tabs.map(({ id, name, query, sourceTable }) => ({ id, name, query, sourceTable })),
```
eugene-mobile, line 431: **`modified` flag is dropped on restore**

The persisted projection only keeps `id`, `name`, `query`, `sourceTable`. A tab that was opened from a SavedQuery and then edited is marked `modified: true` (set on every keystroke at L359), and that flag drives the dirty-dot UI. After restart, the restored tab comes back with `modified` undefined, so the user loses the visual cue that their tab still differs from the saved version — they may overwrite the saved query thinking they were saving fresh edits.

Fix: include `modified` in the projection here and add `modified?: boolean` to `PersistedTab` in `tabPersistence.ts`. (`sourceTable` is already optional; same shape works.)

eugene-mobile (commented):
Implementation is solid and matches the existing per-connection localStorage pattern. One minor UX note inline; otherwise looks good.

PR #155, opened 2026-05-09
feat: create table via UI

### LINKED ISSUE #16: Create table via UI
Today tables can only be created by typing `CREATE TABLE` in the editor.

**Proposal**
- "New table" action (schema-browser or sidebar) opens a designer modal.
- Let users declare columns (name, type, nullability, default), primary key, and indexes.
- Generate `CREATE TABLE` DDL and show it in a preview/confirm step before execution.

eugene-mobile:
Closes #16.

## Summary
A "+" button on the **Tables** group in the schema browser opens a designer modal for composing a \`CREATE TABLE\`. The user fills in column rows (name, type, nullable, default, auto-increment, single PK), watches a live DDL preview, optionally hand-edits it, and submits.

Hidden for MongoDB connections.

## What's in v1
- Schema + table name fields (schema prefilled with the active schema).
- Column rows with: name, type (free-form input + datalist of common types), nullable checkbox, default expression (verbatim — caller quotes strings/dates), auto-increment, PK.
- Single-column PK enforced via radio-style mutex.
- Live DDL preview that's also editable — user edits stop auto-regeneration so manual tweaks aren't blown away.
- "Reset to generated" pill restores form-driven DDL.

## Dialect handling
- **MySQL**: \`AUTO_INCREMENT\`, backtick identifiers.
- **Postgres**: \`GENERATED BY DEFAULT AS IDENTITY\` (modern form, pg10+), double-quote identifiers.

\`/api/connect\` and \`/api/connect/status\` now return \`dbType\` so the frontend knows which dialect to render.

## Out of scope (follow-ups)
- Composite PKs
- Secondary indexes / unique constraints (other than PK)
- Foreign keys, CHECK constraints
- Per-type widgets (VARCHAR length picker, DECIMAL precision/scale, ENUM values)

The hand-editable preview is the escape hatch — anything v1 doesn't model can be typed directly before submitting.

## Implementation
- \`src/lib/sql.ts\` — \`buildCreateTableSql(dialect, schema, table, columns)\`. Forces \`NOT NULL\` on PK and IDENTITY columns regardless of UI state. Strips embedded quote characters from identifiers.
- \`src/components/CreateTableDialog.tsx\` — form + live preview + submit.
- \`src/components/SchemaBrowser.tsx\` — optional \`onCreateTable\` prop renders the "+" trigger on the Tables group row.
- \`src/api.ts\` — \`api.createTable(sql)\`.
- \`server/src/routes/createTable.ts\` — \`POST /api/create-table\` accepts prebuilt SQL, executes via the active driver.

## Test plan
- [x] Unit tests for the DDL helper covering MySQL/Postgres divergence, single PK, NOT NULL forcing, defaults pass-through, identifier-stripping (9 tests).
- [x] Route tests for happy path, missing/empty SQL, driver errors (4 tests).
- [x] \`npm test\` (server: 133, frontend: 41).
- [x] \`npm run build\` clean.
- [ ] Manual in Electron: create a table on MySQL, verify schema browser refreshes; same on Postgres.


src/components/CreateTableDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+        style={{ justifySelf: 'center' }}
+      />
+      <input
+        type="checkbox"
+        checked={col.pk}
+        onChange={(e) => { if (e.target.checked) onSetPk(); else onChange({ pk: false }); }}
+        style={{ justifySelf: 'center' }}
+      />
+      <button onClick={onRemove} style={style.iconBtn} title="Remove column" aria-label="Remove column">×</button>
+      <datalist id="helix-create-table-types">
```
eugene-mobile: **Duplicate `<datalist>` id rendered per row**

`ColumnRow` is rendered once per column, so this `<datalist id="helix-create-table-types">` is emitted N times — every row contributes a duplicate element with the same `id`. That's invalid HTML; browsers will pick one and ignore the rest, which happens to look correct here only because every copy carries the same options.

Lift the `<datalist>` out of `ColumnRow` and render it once at the bottom of the parent `CreateTableDialog`. The `list="helix-create-table-types"` attribute on each input will resolve to that single element.

src/components/CreateTableDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+      if (!c.type.trim()) return `Column "${c.name}" needs a type.`;
+      if (seen.has(c.name)) return `Duplicate column name: ${c.name}`;
+      seen.add(c.name);
+    }
+    return null;
+  };
+
+  const submit = async () => {
+    const v = validate();
+    if (v) { setError(v); return; }
```
eugene-mobile: **Form validation blocks submission of hand-edited SQL**

The PR description calls the editable preview "the escape hatch — anything v1 doesn't model can be typed directly before submitting." But `validate()` runs unconditionally before submit and rejects on form-level state (missing table name, duplicate column names, missing column type). A user who clears the form and writes their own composite-PK / FK / multi-statement DDL into the textarea will be blocked with `"Table name is required."` even though the SQL being submitted is fine.

Skip form validation when the user has overridden the preview:

```ts
if (sqlOverride === null) {
  const v = validate();
  if (v) { setError(v); return; }
}
```

src/lib/sql.ts
```diff
@@ -18,6 +18,58 @@ export function formatSqlValue(v: CellValue): string {
...
+  const autoInc = dialect === 'mysql' ? 'AUTO_INCREMENT' : 'GENERATED BY DEFAULT AS IDENTITY';
+
+  const lines = columns.map(c => {
+    const parts: string[] = [quoteIdent(dialect, c.name), c.type];
+    // PK columns are implicitly NOT NULL across both dialects; Postgres IDENTITY
+    // additionally requires it. Force the constraint on so the DDL stays valid
+    // even if the user ticks PK or auto-inc without un-ticking nullable.
+    if (!c.nullable || c.pk || c.autoIncrement) parts.push('NOT NULL');
+    if (c.default !== null && c.default !== '') parts.push(`DEFAULT ${c.default}`);
+    if (c.autoIncrement) parts.push(autoInc);
```
eugene-mobile, line 65: **Auto-increment without PK produces invalid MySQL DDL**

MySQL requires every `AUTO_INCREMENT` column to be a key (typically the PK). The form lets the user tick `A.I.` without ticking `PK`, and this helper happily emits e.g. `` `id` INT NOT NULL AUTO_INCREMENT `` with no key clause, which MySQL rejects at execution time with *"Incorrect table definition; there can be only one auto column and it must be defined as a key."*

Since this check is cheap and dialect-agnostic, either reject in `validate()` (`"Auto-increment column must also be the primary key."`) or auto-set `pk: true` when `autoIncrement` is checked in `ColumnRow`'s onChange. Postgres `IDENTITY` doesn't have the same constraint, but the same UX rule keeps both dialects consistent.

eugene-mobile (commented):
Nicely scoped v1 with a clean separation between the DDL helper, dialog, and route, and good test coverage on both sides. A few concrete issues below — most notable is that the hand-editable SQL escape hatch is currently blocked by form validation, which undermines the main reason for shipping the textarea.

PR #156, opened 2026-05-09
feat: alter table via UI

### LINKED ISSUE #17: Alter table via UI
No UI today for adding/dropping/renaming columns or adding/dropping indexes.

**Proposal**
- On a selected table, offer "Alter table" — reuse a columns+indexes designer.
- Generate the minimal `ALTER TABLE` statement(s) and preview before execution.
- Support add/drop/rename column, change type, add/drop index, add/drop foreign key.

eugene-mobile:
Closes #17.

## Summary
- Right-click any table in the schema browser → **Edit table** opens the `AlterTableDialog`.
- The dialog pre-loads existing columns (name, type, nullable, default) and lets the user rename columns, change types, toggle nullable, update defaults, add new columns, or mark columns for dropping.
- A live SQL preview (also hand-editable) shows the minimal `ALTER TABLE` statement(s) before submission.
- After a successful alter, the schema browser refreshes automatically.
- Hidden for MongoDB connections (same as Create table).

## Dialect handling
- **MySQL**: one `ALTER TABLE` with `ADD COLUMN`, `DROP COLUMN`, and `CHANGE COLUMN` clauses combined.
- **Postgres**: `RENAME COLUMN` emitted as separate statement(s) (Postgres requirement); type / nullability / default changes combined in one `ALTER TABLE`.
- Hand-edited SQL bypasses form validation (same escape-hatch model as `CreateTableDialog` from #155).

## Out of scope (follow-ups)
- Add/drop indexes and foreign keys — the hand-editable preview is the escape hatch.
- PK and auto-increment changes on existing columns (requires DROP/ADD PRIMARY KEY which is dialect-specific and risky).

## Implementation
- `src/lib/sql.ts` — `AlterTableColumn` type + `buildAlterTableSql()`.
- `src/components/AlterTableDialog.tsx` — form + live preview + submit.
- `src/components/SchemaBrowser.tsx` — "Edit table" context menu item.
- `src/api.ts` — `api.alterTable(sql)`.
- `server/src/routes/alterTable.ts` — `POST /api/alter-table`.
- `server/src/index.ts` — route registration.

## Test plan
- [x] 12 unit tests for `buildAlterTableSql` covering MySQL/Postgres divergence, add/drop/rename/modify/schema-qualified.
- [x] 4 route tests for happy path, missing SQL, whitespace SQL, driver errors.
- [x] `npm test` (server: 137, frontend: 53).
- [x] `npm run build` clean.
- [ ] Manual in Electron: rename a column on MySQL; change type on Postgres; drop a column; add a column; verify schema browser refreshes.


src/lib/sql.ts
```diff
@@ -37,6 +37,107 @@ function quoteIdent(d: SqlDialect, s: string): string {
...
+      } else {
+        const orig = origMap.get(col.originalName);
+        const renamed = col.name !== col.originalName;
+        const changed = orig && (
+          col.type !== orig.type ||
+          col.nullable !== orig.nullable ||
+          (col.default ?? '') !== (orig.default ?? '')
+        );
+        if (renamed || changed) {
+          clauses.push(`CHANGE COLUMN ${q(col.originalName)} ${colDefExpr(dialect, col)}`);
```
eugene-mobile, line 97: **MySQL: `CHANGE COLUMN` re-emits inline `PRIMARY KEY` for already-PK columns**

`colDefExpr` unconditionally appends `PRIMARY KEY` whenever `col.pk` is true. For ADD COLUMN that's fine (assuming the table has no PK yet), but here it's also used for `CHANGE COLUMN` of existing columns. If the user renames or retypes a PK column, the produced statement looks like:

```sql
ALTER TABLE `users`
  CHANGE COLUMN `id` `user_id` BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY;
```

MySQL rejects this with `ERROR 1068 (42000): Multiple primary key defined`, because the inline `PRIMARY KEY` clause is interpreted as adding a new PK constraint, not as a no-op annotation. Since the dialog seeds `pk` from the existing schema and there's no UI to clear it, any rename or type change on a PK column will fail at execute time — and the test suite doesn't cover the PK case, so this slipped through.

Fix: when emitting a `CHANGE COLUMN` for an existing column, pass a variant of `colDefExpr` that omits `PRIMARY KEY` (and probably the inline `AUTO_INCREMENT` is also worth keeping — that one is fine to repeat). Same caveat applies to the Postgres ADD COLUMN path if the user adds a column flagged as PK to a table that already has one, but the current UI doesn't expose a PK toggle for new columns so it's latent.

src/components/AlterTableDialog.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+              return (
+                <>
+                  <input
+                    key={`${i}-name`}
+                    style={s.input}
+                    value={c.name}
+                    onChange={(e) => updateCol(i, { name: e.target.value })}
+                    placeholder="column_name"
+                    readOnly={!isNew && false}
```
eugene-mobile, line 189: **`readOnly={!isNew && false}` is dead code — always `false`**

`!isNew && false` always evaluates to `false`, so this is equivalent to `readOnly={false}` (i.e. the input is always editable). Looks like a leftover from a half-finished thought about restricting renames of existing columns. Either commit to the intent (`readOnly={!isNew}` if name edits should go through a dedicated rename flow) or just delete the prop. As-is it's confusing and likely to trip up the next reader.

src/components/AlterTableDialog.tsx
```diff
@@ -0,0 +1,262 @@
...
+          </div>
+
+          <div style={s.columnsTable}>
+            <div style={s.colHead}>Name</div>
+            <div style={s.colHead}>Type</div>
+            <div style={s.colHead}>Null</div>
+            <div style={s.colHead}>Default</div>
+            <div />
+
+            {cols.map((c, i) => {
```
eugene-mobile, line 160: **Index-based React keys on a mutable list of controlled inputs**

Two issues stacked here:

1. The `<>` shorthand Fragments returned from this `.map` cannot carry a key — React will warn, and more importantly, when items are added/removed/toggled the per-child keys (`${i}-name`, etc.) collide across renders because `i` is the array index, not a stable identity. Removing a new column at index 2, for example, shifts everything after it; React reconciles by index, which means controlled inputs can briefly render with the wrong row's value, focus jumps, and the dropped→active toggle (which swaps `<span>` for `<input>` under the same key) forces unmount/remount.

2. Suggested fix: give each column a stable `id` when it enters state (e.g. `crypto.randomUUID()` for new columns, the `originalName` for existing ones), then `return <Fragment key={col.id}>...</Fragment>` (long-form, since shorthand can't take a key). The inner per-cell keys are then unnecessary.

This matters more than it sounds for a column editor — the user will be typing into these inputs while the list is mutating.

eugene-mobile (commented):
Solid feature scaffold with good test coverage on the dialect-aware SQL builder. Three concerns worth addressing: a real correctness bug for MySQL when altering an existing PK column (CHANGE COLUMN re-emits inline PRIMARY KEY and MySQL rejects it as a duplicate), a dead `!isNew && false` expression that suggests an unfinished thought, and index-based React keys on the column list (combined with shorthand Fragments without keys) which will produce input/focus glitches as columns are added/removed/toggled.

PR #157, opened 2026-05-10
feat: redesign connection manager with auto-detect single-page form

eugene-mobile:
Updates the connection manager to match the design from the [Helix design system handoff bundle](https://api.anthropic.com/v1/design/h/n7jg62UCiunR3I_EtZsatQ) (claude.ai/design).

(First push of this branch shipped a 2-step picker; that was the wrong variation. Force-pushed to the auto-detect single-page form.)

## What you see now
- **Header** — title + DetectionBadge pill that shows _Likely MySQL_ (warning, predicted from URI scheme or port) → _Identifying…_ (spinner during connect) → _MySQL · SQL_ (accent, after a successful Test).
- **Connection name** — input with a chevron dropdown listing saved connections. Focus shows all; typing filters. Click loads the connection (and pulls the keychain password on Electron).
- **Dual input** — Connection string box AND Host/Port/User/Password/Database box, both visible. The last-edited side gets the accent border + glow. Edits on either side sync to the other; the URI scheme follows the predicted type so it auto-switches to \`postgresql://\` or \`mongodb://\` as the port changes.
- **Password masking** — when the user is editing fields, the rendered URI shows \`***\` instead of the real password so it's never visible in the connection-string box.
- **SSL toggle** — plus the existing "Verify server certificate" sub-option with the unverified warning.

## What's preserved
- Save-password keychain integration on Electron.
- Test-connection button + success/error banners.
- MongoDB URI mode (now just "type into the connection-string box"; \`mongoMode\` stays in the form contract for backwards compat with saved entries).
- The \`ConnectionForm\` shape — \`App.tsx\` and the backend contract didn't change.

## What's out of scope
- The animated URI ticker placeholder from the prototype (cute but non-essential — used a static placeholder string instead).
- The "Open in Helix" confirmed-state CTA — the existing app advances the user past the modal as soon as Connect succeeds, so the confirmed state only appears after Test.
- Restyled \`TopBar\`, \`SchemaBrowser\`, \`QueryEditor\`, \`ResultsTable\`, multi-DB schema browser — separate redesigns; this PR only touches the connection manager.

## Test plan
- [x] \`npm test\` (frontend: 41, server: 133) — all green.
- [x] \`npm run build\` clean.
- [ ] Manual in Electron: confirm modal layout matches mock; confirm name dropdown shows saved entries; confirm typing in the URI populates host/port/user/password/db; confirm typing in fields rebuilds the URI; confirm SSL verify sub-option still surfaces the warning; confirm keychain password loads on saved-connection click.

### DIFF
--- a/src/components/ConnectionManager.tsx
+++ b/src/components/ConnectionManager.tsx
@@ -1,5 +1,5 @@
-import { useState, useEffect } from 'react';
-import type { CSSProperties, ReactNode } from 'react';
+import { useState, useEffect, useMemo, useRef } from 'react';
+import type { CSSProperties } from 'react';
 import type { Theme } from '../theme';
 import { api } from '../api';
 import { listSavedConnections, deleteSavedConnection, type SavedConnection } from '../savedConnections';
@@ -20,10 +20,10 @@ export interface ConnectionForm {
   // Only meaningful in Electron — when true, the password is persisted via
   // safeStorage and reloaded the next time this connection is opened.
   savePassword: boolean;
-  // Only meaningful when type === 'mongodb' AND mongoMode === 'uri'.
+  // Only meaningful when type === 'mongodb' AND the user is editing via URI.
   connectionString?: string;
-  // UI state: which input mode the mongodb form is in. Persisted alongside
-  // the connection itself (derived on load: presence of connectionString).
+  // UI state: which input mode is active for the mongodb form. Persisted on
+  // saved connections; in this redesign it's also driven by `lastEdited`.
   mongoMode?: 'fields' | 'uri';
 }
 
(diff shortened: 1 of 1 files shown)


src/components/ConnectionManager.tsx
```diff
@@ -565,3 +403,375 @@ export function ConnectionManager({ onConnect, isConnecting, error, t }: Connect
...
+  onSelect: () => void;
+  onDelete: (e: React.MouseEvent) => void;
+  t: Theme;
+}) {
+  const [hovered, setHovered] = useState(false);
+  const active = hovered || highlighted;
+  return (
+    <div
+      onClick={onSelect}
+      onMouseEnter={() => setHovered(true)}
```
eugene-mobile, line 598: **Escape key swallowed even when the dropdown is closed**

The `onKey` handler reaches this branch and calls `e.nativeEvent.stopImmediatePropagation()` regardless of whether `open` is `true` or `false`. So when the user focuses the Connection Name field (with no dropdown showing) and presses Escape to dismiss the connection modal, the event is eaten here and the modal never closes.

The old code had an early-return guard:
```ts
if (!suggestOpen || filteredSaved.length === 0) return;
```
that prevented this. The fix is minimal:
```ts
else if (e.key === 'Escape' && open) {
```

eugene-mobile (commented):
Solid redesign overall — the single-page layout with live URI/field sync is much cleaner, and the extracted `ConnectionNameField` and `DetectionBadge` components are well-structured. One regression: the Escape handler in `ConnectionNameField` swallows the key unconditionally, which breaks the parent dialog's close behaviour.

src/components/ConnectionManager.tsx
eugene-mobile: Good catch — fixed in d7a5102. Scoped the branch to `open === true` so we only `stopImmediatePropagation` when there's actually a dropdown to dismiss.

PR #161, opened 2026-05-14
fix: pick free server port and surface startup failures in packaged app

eugene-mobile:
## Symptom
A user installing Helix from the DMG and launching it saw a blank window
with **"Cannot GET /"** — Express's default 404 — instead of the app.

## Root cause (two issues compounding)
**1. Hardcoded port + no conflict handling.** \`electron/main.cjs\` told
the bundled server to bind \`:3001\`. If anything else was already
listening — most commonly a leftover dev \`tsx watch\` from a prior
debugging session — \`app.listen\` failed with \`EADDRINUSE\` and the
utilityProcess child crashed. But the BrowserWindow then loaded
\`http://localhost:3001/\` against whatever **else** happened to be on
3001. That other listener typically has no \`STATIC_PATH\` (only the
bundled server sets it via env) so it has no SPA fallback route → "Cannot
GET /".

**2. Silent startup failure.** \`waitForServer\` only resolved on
success; the prior \`startServer\` had no \`serverProc.on('exit')\`
hook. A child that died during startup was indistinguishable from "still
booting" until the 50-attempt timeout, and even then the user got a
generic timeout message with no captured stderr.

## Fix
- **\`findFreePort()\`** — uses \`net.createServer().listen(0)\` to ask
  the OS for an unused port, closes it, returns the number. A small race
  window remains (something could grab it between close and fork) but
  in practice this is many orders of magnitude better than colliding
  with a known stale listener.
- **\`startServer\` is now async** — captures the child's stderr into a
  buffer, registers \`on('exit')\`, and rejects with the captured stderr
  (or a fallback message) if the child dies before readiness probing
  succeeds.
- **\`app.whenReady\` catches the rejection** and surfaces a
  \`dialog.showErrorBox\` with the underlying cause, then quits. No more
  silent fallthrough into a BrowserWindow loading nothing.
- \`PORT\` / \`APP_URL\` are now \`let\` so the dynamic port reaches
  \`win.loadURL\` via closure.

Dev mode is untouched: the dev server is started outside this process by
\`concurrently\`, and the BrowserWindow loads the Vite URL.

## Test plan
- [x] Packaged \`Helix.app\` boots, picks a free port (verified
  \`http://localhost:62349/\` returned HTTP 200 with the bundled
  \`index.html\`).
- [x] No regression in dev mode (no code paths in dev branch changed).
- [x] Error-handling wiring reviewed by inspection — \`on('exit')\`
  rejects, the outer \`await\` shows the dialog, app quits.
- [ ] Manual: install fresh DMG (after merge + rebuild) on a clean Mac,
  verify no "Cannot GET /" regression.

### DIFF
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -3,14 +3,20 @@
 const { app, BrowserWindow, Tray, Menu, nativeImage, utilityProcess, dialog, ipcMain, safeStorage, powerMonitor } = require('electron');
 const path = require('path');
 const http = require('http');
+const net = require('net');
 const fs = require('fs');
 
 const isDev = !app.isPackaged;
-const PORT = 3001;
+// In dev the server is started by `npm run dev:server` outside this process,
+// and is hardcoded to 3001. In production we pick a free port at startup so
+// a leftover dev server (or any other listener) can't shadow us — the prior
+// behaviour silently failed to bind, leaving the BrowserWindow to load
+// whatever happened to be answering on 3001 (often "Cannot GET /").
+let PORT = 3001;
+let APP_URL = `http://localhost:${PORT}`;
 
 app.setName('Helix');
 const VITE_URL = 'http://localhost:5173';
-const APP_URL = `http://localhost:${PORT}`;
 const ERR_CONNECTION_REFUSED = -102;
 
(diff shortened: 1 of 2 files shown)


eugene-mobile:
@Claude review this PR

PR #170, opened 2026-05-23
feat: classify driver errors into client/transient/server for accurate HTTP status codes

### LINKED ISSUE #122: Classify driver errors into 4xx vs 5xx in query route
## Context

In `server/src/routes/query.ts`, all `driver.query()` rejections currently map to **400**:

```ts
} catch (err) {
  const message = err instanceof Error ? err.message : String(err);
  res.status(400).json({ error: message });
}
```

This was preserved from the pre-existing SQL behaviour (see PR #121 / issue #109 follow-up review thread). For SQL it was already lossy — a connection drop or timeout is reported as a client error. With the MongoDB driver landed, the surface broadens further: `MongoServerSelectionError`, `MongoNetworkError`, and connect-time failures inside `ensureConnected()` are infrastructure problems, not client mistakes, and they will now be reported as 400.

## Proposal

Have each driver expose a structured error class on thrown errors so the route can classify, e.g.:

```ts
type DriverErrorClass = 'client' | 'transient' | 'server';
class DriverError extends Error {
  errorClass: DriverErrorClass;
  // ...
}
```

Then in the route:

- `client` -> 400 (parse / unsupported operation / validation)
- `transient` -> 503 (connection / timeout / retryable)
- `server` -> 500 (internal)

Both `PostgresDriver` (or future SQL drivers) and `MongoDBDriver` would need to wrap or re-throw with the appropriate class. The code comment near the `catch` block should be removed once this lands.

## Notes

- Pre-existing for the SQL path; MongoDB just made the surface more visible.
- No security impact — purely error-message / status-code accuracy.
- Best done after both drivers expose structured error codes (a few existing `throw new Error(...)` sites would need to become `throw new DriverError(..., 'client')`).

eugene-mobile:
Implements #122: introduces DriverError with errorClass so the query route returns 400/503/500 instead of always 400.

Changes:
- interface.ts: DriverErrorClass + DriverError class
- postgres.ts: classifyPgError() maps pg error codes to classes
- mongodb.ts: explicit throws converted to DriverError, Mongo errors classified by type
- query.ts: catch block maps errorClass to HTTP status
- Tests updated and extended

Closes #122

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/server/src/drivers/interface.ts
+++ b/server/src/drivers/interface.ts
@@ -1,3 +1,14 @@
+export type DriverErrorClass = 'client' | 'transient' | 'server';
+
+export class DriverError extends Error {
+  readonly errorClass: DriverErrorClass;
+  constructor(message: string, errorClass: DriverErrorClass, options?: ErrorOptions) {
+    super(message, options);
+    this.name = 'DriverError';
+    this.errorClass = errorClass;
+  }
+}
+
 export interface ConnectionConfig {
   host: string;
   port: number;
--- a/server/src/drivers/mongodb.test.ts
+++ b/server/src/drivers/mongodb.test.ts
@@ -62,10 +62,14 @@ const { ObjectIdCtor } = vi.hoisted(() => {
 vi.mock('mongodb', () => ({
   MongoClient: vi.fn(() => mockClient),
   ObjectId: ObjectIdCtor,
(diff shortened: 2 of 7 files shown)


eugene-mobile:
@claude review this PR

PR #174, opened 2026-06-12
feat: autocomplete column / table / keyword names in the SQL editor

### LINKED ISSUE #171: Suggestions on field names in SQL entry input
When user is editing a SQL query, we should be able to suggest field names as we know table schema.

eugene-mobile:
Closes #171.

## Summary
- The query editor now offers a context-aware suggestion popup as the user types. Anchors at the caret using a textarea-mirror to measure caret coordinates, so it tracks line / column wrapping, scroll, padding, and font without needing a heavier editor like Monaco.
- The popup classifies the slot at the caret before deciding what to suggest. Each branch is intentionally restrictive so the list reads "useful" rather than "everything in the schema":
  - **Start of statement / after `;`** → statement-starter keywords only (`SELECT`, `INSERT INTO`, `UPDATE`, `DELETE FROM`, `CREATE TABLE`, `ALTER TABLE`, `DROP TABLE`, `EXPLAIN`, `WITH`, `USE`, `BEGIN`, `COMMIT`, …).
  - **After `FROM` / `JOIN` / `INTO` / `UPDATE` / `TRUNCATE` / `DESCRIBE`** → table names only.
  - **After `WHERE` / `SELECT` / `ON` / `AND` / `OR` / `HAVING` / `,` / `=` / `<` / `>` / `(` / etc.** → columns of every table in the `FROM`/`JOIN` scope, mixed with expression keywords (`AND`, `OR`, `IS NULL`, `IN`, `LIKE`, `BETWEEN`, `DISTINCT`, `NULL`, `TRUE`, `FALSE`, `CASE`/`WHEN`/`THEN`/`ELSE`/`END`).
  - **After `GROUP` / `ORDER`** → `BY` only.
  - **Qualified prefix (`u.col`)** → columns of the table that `u` resolves to (alias-first, table-name fallback), bypassing slot inspection.
  - **Permissive fallback** (e.g. typing `wh` on a new line after `FROM \`tbl\``) → clause keywords like `WHERE`, `GROUP BY`, `ORDER BY`, `HAVING`, `LIMIT`, `JOIN`.
- Aliases are surfaced too — typing `o` after `FROM users u JOIN orders o ON …` shows `o` as an alias hint with its underlying table.
- Keyword rows are coloured with the editor's `sqlKeyword` theme token and rendered bold so they stand out from column/table identifiers. Each suggestion kind also has its own icon (square brackets for keywords, table grid, columns, list-lines for aliases).
- Keyboard: `↑ / ↓` navigate, `Tab` or `Enter` accept, `Esc` closes. Mouse hover updates the highlight; click accepts. Caret-moving keys (arrows / Home / End) and `mousedown` inside the textarea close the popup so it doesn't drift off a stale word. `onBlur` closes with a small delay so popup clicks still register.
- The FROM/JOIN parser intentionally **masks the prefix region being typed** before scanning refs. Without that, a partially-typed identifier directly after the FROM clause would get captured as an alias (so typing `wh` on a new line after `FROM users` previously made the parser think `wh` is the alias of `users`).
- MQL connections skip autocomplete entirely.

## Test plan
- [x] `npx tsc -b` — clean.
- [x] `npm test` — 76/76 pass (23 new tests in \`src/lib/sqlContext.test.ts\` covering the FROM parser, the caret-context detector, and the slot classifier; existing 53 untouched).
- [x] \`npm run build\` — succeeds.
- [x] Manual against a real MySQL:
  - Typing \`SEL\` at the top of a query → \`SELECT\` is the first item.
  - Typing \`us\` after \`FROM \` → table list filtered to those containing \`us\`.
  - Typing \`u.\` after \`FROM users u\` → user columns only.
  - Typing \`wh\` on a new line after \`FROM \`fi_transaction\`\` → \`WHERE\` leads the list, no spurious alias suggestion.

### DIFF
--- a/src/components/QueryEditor.tsx
+++ b/src/components/QueryEditor.tsx
@@ -5,6 +5,9 @@ import type { Theme } from '../theme';
 import type { SchemaData } from '../api';
 import type { HistoryEntry } from '../queryHistory';
 import type { SavedQuery } from '../savedQueries';
+import { SuggestionPopup, type Suggestion } from './SuggestionPopup';
+import { extractTableRefs, contextAtCaret, expectedSlot, KEYWORD_GROUPS } from '../lib/sqlContext';
+import { getCaretCoordinates } from '../lib/caretPosition';
 
 const LARGE_ROW_THRESHOLD = 5_000;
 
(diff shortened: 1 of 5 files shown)


eugene-mobile:
@claude review this PR

PR #176, opened 2026-06-13
fix: keep query editor toolbar visible with large queries (#173)

### LINKED ISSUE #173: Incorrect scrolling with big queries
If a query is big, tab with "Execute" and other buttons are being scrolled away: 

Image

Users should be able to access actions even if they are at the bottom of the query.

eugene-mobile:
## Problem

Closes #173.

When a query is tall, the editor toolbar (Run / Explain / Format buttons) scrolls out of view, so users can't trigger actions while at the bottom of a large query.

issue

## Root cause

`QueryEditor`'s root container is a flex column but had **no height constraint**. Its parent in `App.tsx` is a fixed `flex: 0 0 240px` box with `overflow: hidden`. Because the root defaulted to `flex: 0 1 auto` (and flex items have `min-height: auto`, which prevents shrinking below content), a tall query grew the whole editor past 240px. Focusing the textarea near the bottom then made the browser scroll the `overflow:hidden` parent to reveal the caret — dragging the pinned toolbar off the top of the viewport. The textarea's own `overflowY: auto` never engaged because the textarea was never height-bounded.

## Fix

- Add `flex: 1; min-height: 0` to the editor root so it fills its fixed-height parent. The bounded `editorWrap` now constrains the textarea, whose existing `overflowY: auto` scrolls the query **internally** while the toolbar (`flex-shrink: 0`) stays pinned and always reachable.
- Sync the line-number gutter's `scrollTop` to the textarea on scroll. Previously the gutter and text scrolled together as one unit via the parent; now that the textarea scrolls on its own, this keeps line numbers aligned with the text.

## Testing

- `tsc --noEmit` passes; no new lint errors in the edited file; all 76 unit tests pass.
- Manual: paste a long query, scroll to the bottom — the Run/Explain toolbar stays fixed at the top and line numbers remain aligned.

### DIFF
--- a/src/components/QueryEditor.tsx
+++ b/src/components/QueryEditor.tsx
@@ -86,6 +86,7 @@ export function QueryEditor({
 }: QueryEditorProps) {
   const isMql = queryMode === 'mql';
   const textareaRef = useRef<HTMLTextAreaElement>(null);
+  const lineNumsRef = useRef<HTMLDivElement>(null);
 
   const [historyOpen, setHistoryOpen] = useState(false);
   const historyAnchorRef = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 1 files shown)


eugene-mobile:
@claude review this PR

PR #177, opened 2026-06-14
feat: per-query timeout that cancels over-running queries (#22)

### LINKED ISSUE #22: Query timeout setting
Long-running queries can't be cancelled today.

**Proposal**
- Add a per-connection (or per-query) timeout.
- On timeout, run `KILL QUERY` against the offending thread id.
- Surface the reason in the results error banner.

eugene-mobile:
## Problem

Closes #22. Long-running queries couldn't be cancelled — once a slow query was running there was no way to abort it from the UI.

## What this does

Adds a user-configurable **per-query timeout**, shown as a compact picker in the query editor toolbar (No timeout / 10s / 30s / 1m / 5m) and persisted in `localStorage` (default **30s**).

The client sends `timeoutMs` with each `/api/query` request. When the deadline passes, the driver cancels the statement **on the server** using its idiomatic mechanism, and rejects with a `transient` `DriverError` whose message — `Query cancelled: exceeded the Ns timeout.` — surfaces in the existing results error banner (HTTP 503).

| Driver | Cancellation mechanism |
|--------|------------------------|
| MySQL | `KILL QUERY <threadId>` on a separate pooled connection |
| Postgres | `SELECT pg_cancel_backend(<pid>)` on a separate pooled client |
| MongoDB | `maxTimeMS` on the read ops (`find`/`findOne`/`aggregate`/`count`) |

A separate connection is used for the SQL cancel because the one running the timed-out query is busy.

### Connection hygiene on timeout

On timeout the SQL drivers **discard** the connection (`mysql2` `destroy()` / `pg` `release(true)`) rather than returning it to the pool: the `KILL`ed/cancelled statement is still draining on that socket, and a reused connection could hand the next caller a half-read result set. The pool opens a fresh connection on the next request.

### No unhandled rejections

A shared `withQueryTimeout` helper races the work promise against the deadline. `Promise.race` keeps a handler attached to the work promise, so the statement's eventual rejection (arriving just after the cancel) is consumed and never surfaces as an unhandled rejection. A falsy/zero/negative timeout disables the deadline entirely, so internal callers (insert/update/delete/explain) are unaffected.

## Scope

The timeout applies only to the user-facing `/api/query` path. Internal CRUD/DDL routes call the drivers without a timeout (unchanged 3-arg behavior).

## Testing

- New `server/src/drivers/timeout.test.ts` (13 tests): label/error formatting, work-wins, no-timeout passthrough, work-rejection passthrough, deadline-wins fires `onTimeout` + transient error, and the late-rejection-is-not-unhandled case (fake timers).
- New route tests: `timeoutMs` forwarding to `query`/`queryAll`, disarming on absent/0/negative/non-numeric, and 503 mapping for a transient timeout error.
- Updated existing route + mongodb assertions for the new argument shape.
- All suites green: **182 server / 76 client**. Client `tsc` + production build clean.

> [!NOTE]
> I could not run a full live end-to-end check (slow query → cancel) — the backend on this machine had no active DB connection and the browser-automation extension wasn't connected. The logic is covered by unit tests; a manual `SELECT SLEEP(10)` with a 5s timeout against a real MySQL/Postgres is worth a smoke test before release.

### DIFF
--- a/server/src/drivers/interface.ts
+++ b/server/src/drivers/interface.ts
@@ -84,13 +84,19 @@ export interface DbDriver {
    * accept a MQL request object. Routes branch on this to validate the body shape.
    */
   readonly queryMode: 'sql' | 'mql';
-  /** Run a query, optionally switching schema first (atomically on one connection). */
-  query(sql: string, params?: unknown[], schema?: string): Promise<QueryResult>;
+  /**
+   * Run a query, optionally switching schema first (atomically on one connection).
+   * When `timeoutMs` is set the driver cancels the statement on the server after
+   * that many ms (MySQL `KILL QUERY`, Postgres `pg_cancel_backend`, MongoDB
+   * `maxTimeMS`) and rejects with a `transient` DriverError.
+   */
+  query(sql: string, params?: unknown[], schema?: string, timeoutMs?: number): Promise<QueryResult>;
   /**
    * Run one or more semicolon-separated statements and return a result per statement.
-   * Implemented by sql-mode drivers; mql-mode drivers can omit it.
+   * Implemented by sql-mode drivers; mql-mode drivers can omit it. `timeoutMs` has
+   * the same cancel-on-deadline semantics as {@link query}.
    */
-  queryAll?(sql: string, schema?: string): Promise<QueryResult[]>;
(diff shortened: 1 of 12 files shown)


eugene-mobile:
@claude  review this PR

PR #178, opened 2026-06-14
fix: single-click on sidebar table only toggles columns, not a new tab (#54)

### LINKED ISSUE #54: Single-click on table in sidebar unexpectedly opens a new query tab
## Problem
Single-clicking a table name in the sidebar opens a new query tab pre-filled with `SELECT * FROM table LIMIT 100;` and immediately switches to it.

Users who simply want to expand the tree to view column names are unintentionally creating new tabs and losing their current editor context.

## Steps to reproduce
1. Have a query open in the editor with unsaved work
2. Single-click any table name in the sidebar
3. A new tab is created and the user is navigated away from their current work

## Expected behaviour
- Single-click should toggle the column list expand/collapse only
- Opening a new query tab for a table should be a deliberate action: double-click, or a dedicated "Query table" button/context menu item
- The sidebar expand/collapse should not affect the current active tab

## Notes
The current auto-generated query uses backtick-quoted table names and adds LIMIT 100, which is a reasonable template — the issue is just that single-click triggers it unexpectedly.

eugene-mobile:
## Problem

Closes #54. Single-clicking a table name in the schema browser both expanded/collapsed its column list **and** opened (or switched to) a `SELECT * FROM table LIMIT 100;` query tab. Users who just wanted to peek at column names were unintentionally navigated away from their current editor work.

## Fix

In `SchemaBrowser.tsx`, the table row's click handling is split:

- **Single-click** → toggles the column list expand/collapse only (`toggleTable`). It no longer touches the active tab or `onTableSelect`.
- **Double-click** → opens the query tab (`onTableSelect`), the deliberate action.
- **Right-click → "Query table"** → a new context-menu item, same deliberate `onTableSelect` path, for discoverability.

The hover tooltip now hints "double-click to query".

`onTableSelect` itself is unchanged, so the existing behaviour it drives is preserved: dedupe by `(schema, table)` so re-querying a table reuses its tab, and the PK-aware `SELECT` template (descending sort only for a single auto-increment PK).

## Notes

- Double-clicking fires two `onClick`s (toggle → toggle) before `onDoubleClick`, so expansion nets to its prior state while the tab opens — the standard tree-view trade-off, visually negligible.
- No behaviour test added: the repo has no component-test infra (no jsdom/RTL, vitest only matches `.test.ts`) and no existing SchemaBrowser tests. The change is pure handler rewiring; typecheck, lint, and the production build pass, and the existing 76 client tests are green.

> Per request, the app was not launched for manual verification.

### DIFF
--- a/src/components/SchemaBrowser.tsx
+++ b/src/components/SchemaBrowser.tsx
@@ -250,12 +250,17 @@ export function SchemaBrowser({ schema, activeTable, onTableSelect, onSchemaChan
               <div key={table.name}>
                 <div
                   style={{ ...s.tableRow, ...(activeTable === table.name ? s.tableRowActive : {}) }}
-                  onClick={() => { onTableSelect(table.name); toggleTable(table.name); }}
+                  // Single-click only expands/collapses the column list — it must
+                  // not steal the user's current editor tab. Opening a query tab
+                  // is a deliberate action: double-click, or the context menu's
+                  // "Query table". See #54.
+                  onClick={() => toggleTable(table.name)}
+                  onDoubleClick={() => onTableSelect(table.name)}
                   onContextMenu={(e) => {
                     e.preventDefault();
                     setContextMenu({ x: e.clientX, y: e.clientY, name: table.name, type: 'table' });
                   }}
-                  title={table.comment || undefined}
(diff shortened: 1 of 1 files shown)


eugene-mobile:
@claude review this PR

PR #181, opened 2026-06-19
fix: driver-aware MCP toolset so MongoDB connections work (#180)

### LINKED ISSUE #180: MCP server is unusable on MongoDB connections (SQL-only tools block all MQL queries)
## Summary

When Helix is connected to a **MongoDB** database, the MCP server is effectively unusable for AI agents — no query can return a single row. The MCP tool surface is hard-coded to be **SQL-shaped** regardless of the connected driver, and the SQL validation layer and the MongoDB driver impose mutually-exclusive requirements.

## Root cause

`server/src/mcp.ts` always registers the same SQL-oriented tools (`execute_query`, `execute_write`, `list_tables`, `describe_table`), even when the active driver is `MongoDBDriver` (`queryMode: 'mql'`, see `server/src/drivers/mongodb.ts`):

1. **The tool descriptions tell the model to write SQL.** `execute_query` says *"Run a read-only SQL query (SELECT / SHOW / DESCRIBE / EXPLAIN / WITH)"* and its only input is a `sql` string. But the Mongo driver expects a **JSON-encoded MQL request** (`{collection, operation, filter, pipeline, ...}`).
2. **The two validation layers contradict each other:**
   - SQL in → passes the `firstKeyword()` keyword gate (`SELECT`/`SHOW` are whitelisted) → reaches the Mongo driver → driver rejects with *"MongoDB driver expects a JSON-encoded MQL request."*
   - MQL JSON in → `firstKeyword()` runs first; `{"collection"...` matches no keyword → rejected with *"Got: unknown"*; `db.coll.find(...)` → matches `db` → *"Got: DB."*

   So neither layer can ever be satisfied on a Mongo connection.

## Observed impact

An agent investigating a production signup-dedup incident burned through ~8 query variants (plain SQL, `SELECT COUNT`, `SHOW TABLES`, JSON `find`, JSON `aggregate`, `db.coll.find` shell syntax) — every one blocked. It correctly concluded the tool was broken for the connection and gave up, unable to run a read it should have been able to run. `describe_table` works (so connectivity/schema are fine); only the query path is broken.

## Proposed fix

Make the MCP toolset **driver-aware**. In `buildMcpServer()`, branch on `getDriver().queryMode` and register one of two mutually-exclusive tool sets:

- **SQL connection** → keep `execute_query` / `execute_write` / `list_tables` / `describe_table` as-is.
- **MongoDB connection** → native typed tools so the model fills structured params instead of hand-authoring MQL JSON:
  - reads: `find_documents`, `aggregate_documents`, `count_documents`, `list_collections`, `describe_collection` (the last also surfaces indexes + JSON Schema validator)
  - writes (gated by "Allow MCP to modify data"): `insert_document`, `update_document`, `delete_document`

Additionally, make the MCP `instructions` **mode-aware** so the agent knows upfront whether it's talking to a SQL or NoSQL database and never attempts the wrong query language.

This is enabled cleanly by the existing architecture: the transport is stateless and `buildMcpServer()` already runs per request, so `tools/list` reflects the currently-connected database.

### Known follow-ups (out of scope for the first fix)
- `_id` ObjectId matching in `find`/`count` filters (string `_id` is matched literally; relates to existing TODO #107).
- Stateless transport means a client that cached the tool list won't get a `tools/list_changed` push if the user switches DB type mid-session.

eugene-mobile:
Fixes #180.

## Problem

When Helix is connected to MongoDB, the MCP server is unusable for AI agents — no query returns a single row. The tools were hard-coded to be SQL-shaped regardless of the connected driver, and the SQL keyword gate (`firstKeyword`) and the MQL driver impose mutually-exclusive requirements:

- SQL in → passes the keyword gate → driver rejects with *"expects a JSON-encoded MQL request"*
- MQL JSON in → rejected by the keyword gate (*"Got: unknown"* / *"Got: DB."*) before reaching the driver

See #180 for the full write-up and an agent transcript hitting the dead-end.

## Fix

`buildMcpServer()` now branches on `getDriver().queryMode` and registers one of two mutually-exclusive tool sets:

| mode | tools |
|------|-------|
| `sql` | `execute_query`, `execute_write`, `list_tables`, `describe_table` (unchanged) |
| `mql` | reads: `find_documents`, `aggregate_documents`, `count_documents`, `list_collections`, `describe_collection` · writes (gated): `insert_document`, `update_document`, `delete_document` |

The MongoDB tools expose MQL fields (`filter`, `projection`, `sort`, `pipeline`, `document`, `update`, …) as **typed parameters** and assemble the JSON-encoded request the driver already parses — the model never hand-authors MQL JSON. `describe_collection` also surfaces indexes + the JSON Schema validator via the driver's existing `getCollectionInfo()`.

**Agents know SQL vs NoSQL upfront** two ways: (1) the wrong tools aren't registered (an agent on Mongo never sees `execute_query`), and (2) the MCP `instructions` are mode-aware and name the connected DB type, e.g. *"This is a MongoDB (document) database… Do NOT write SQL."*

This slots into the existing architecture cleanly: the transport is stateless and `buildMcpServer()` already runs per request, so `tools/list` reflects the currently-connected database.

## Tests

New `server/src/mcp.test.ts` (7 tests, via the SDK in-memory transport):
- Mongo connection exposes Mongo tools and **not** `execute_query`/`execute_write`
- SQL connection exposes SQL tools and not the Mongo tools
- mode-aware instructions say MongoDB / don't-write-SQL
- `find_documents` assembles the correct `{collection, operation:'find', filter, limit}` JSON and passes schema through
- `insert_document` blocked when writes disabled, allowed when enabled
- `update_document` requires a `filter` or `id`

All 169 server tests pass; `mcp.ts` and `mcp.test.ts` are type-clean and `mcp.ts` is lint-clean. (3 pre-existing `tsc` errors in unrelated driver test files are untouched.)

## Known follow-ups (out of scope)
- `_id` ObjectId matching in `find`/`count` filters (string `_id` matched literally; relates to existing TODO #107). An `id` convenience param is exposed on update/delete.
- Stateless transport: a client that cached the tool list won't get a `tools/list_changed` push if the user switches DB type mid-session.

### DIFF
--- a/server/src/mcp.test.ts
+++ b/server/src/mcp.test.ts
@@ -1,0 +1,156 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+vi.mock('./db.js', () => ({
+  getDriver: vi.fn(),
+  getActiveConfig: vi.fn(),
+  isConnected: vi.fn(),
+}));
+vi.mock('./mcp-state.js', () => ({
+  isMcpWritesAllowed: vi.fn(),
+}));
+
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js';
+import { getDriver, getActiveConfig, isConnected } from './db.js';
+import { isMcpWritesAllowed } from './mcp-state.js';
+import { buildMcpServer } from './mcp.js';
+
+async function connectClient() {
+  const server = buildMcpServer();
+  const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
+  const client = new Client({ name: 'test', version: '0' });
+  await Promise.all([server.connect(serverTransport), client.connect(clientTransport)]);
+  return { client, server };
+}
+
+function mockMongoDriver(query = vi.fn()) {
+  const driver = {
+    queryMode: 'mql' as const,
+    query,
+    getSchemas: vi.fn().mockResolvedValue(['shop']),
+    getSchema: vi.fn().mockResolvedValue({ tables: [{ name: 'users', rows: 3 }], views: [] }),
+    getTable: vi.fn().mockResolvedValue({ name: 'users', rows: 3, columns: [] }),
+    getCollectionInfo: vi.fn().mockResolvedValue({ validator: null, indexes: [] }),
+  };
+  vi.mocked(getDriver).mockReturnValue(driver as any);
+  return driver;
+}
+
+describe('MCP server – driver-aware toolset', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    vi.mocked(isConnected).mockReturnValue(true);
+    vi.mocked(getActiveConfig).mockReturnValue({ type: 'mongodb', database: 'shop' } as any);
+    vi.mocked(isMcpWritesAllowed).mockReturnValue(false);
+  });
+
+  it('exposes MongoDB tools (not SQL tools) on a Mongo connection', async () => {
+    mockMongoDriver();
+    const { client } = await connectClient();
+    const names = (await client.listTools()).tools.map(t => t.name).sort();
+
+    expect(names).toContain('find_documents');
+    expect(names).toContain('aggregate_documents');
+    expect(names).toContain('count_documents');
+    expect(names).toContain('list_collections');
+    expect(names).not.toContain('execute_query');
+    expect(names).not.toContain('execute_write');
+  });
+
+  it('exposes SQL tools on a SQL connection', async () => {
+    vi.mocked(getActiveConfig).mockReturnValue({ type: 'postgres', database: 'shop' } as any);
+    vi.mocked(getDriver).mockReturnValue({ queryMode: 'sql' } as any);
+    const { client } = await connectClient();
+    const names = (await client.listTools()).tools.map(t => t.name).sort();
+
+    expect(names).toContain('execute_query');
+    expect(names).toContain('execute_write');
+    expect(names).not.toContain('find_documents');
+  });
+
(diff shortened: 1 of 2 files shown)


eugene-mobile:
@claude review this PR

eugene-mobile:
## Follow-up: handle DB connect/switch mid-session

Addressing review feedback — the original per-mode registration had a real gap, not a minor one: an agent that connects its MCP client **before** a database is selected (the common case for a troubleshooting session) caches `tools/list`. The transport is stateless (`sessionIdGenerator: undefined`), so the server can't push `tools/list_changed`. After the user then connects MongoDB, the agent still holds the SQL tools, and calling `execute_query` hits a server that no longer registers it → a cryptic *"tool not found"* with no hint why.

### Fix: invariant tool list + per-call mode guards

The tool list is now **the same regardless of connected DB**, so there is no stale cache to invalidate:

- **All tools always registered** — both families plus a new `connection_info` tool. `tools/list` never changes with DB type.
- **Per-call mode guard** (`requireMode`) — a tool from the wrong family returns an *actionable* error ("This MCP connection is a MongoDB database… use find_documents/…; call connection_info / re-list / reconnect") instead of "tool not found". The agent self-corrects.
- **`connection_info`** — always available; returns the connected DB type, query language, write state, and recommended tools, so the agent orients up front and can detect a mid-session switch.
- **Mode-aware instructions** still name the currently-connected DB and point at `connection_info`.

This is client-agnostic (doesn't depend on a client honoring `tools/list_changed`) and self-correcting. `mcp.test.ts` now asserts the invariant toolset, `connection_info` output, and that cross-family calls return actionable mismatch errors. All 171 server tests pass.

PR #182, opened 2026-06-19
fix: MongoDB support for the MCP server (driver-aware tools + connection awareness) (#180)

### LINKED ISSUE #180: MCP server is unusable on MongoDB connections (SQL-only tools block all MQL queries)

eugene-mobile:
Fixes #180. (Supersedes #181.)

## Problem

When Helix is connected to MongoDB, the MCP server is unusable for AI agents — no query returns a single row. The tools were hard-coded to be SQL-shaped regardless of the connected driver, and the SQL keyword gate (`firstKeyword`) and the MQL driver impose mutually-exclusive requirements:

- SQL in → passes the keyword gate → driver rejects with *"expects a JSON-encoded MQL request"*
- MQL JSON in → rejected by the keyword gate (*"Got: unknown"* / *"Got: DB."*) before reaching the driver

See #180 for the full write-up and an agent transcript hitting the dead-end.

## Fix

The MCP tool list is now **invariant** across connections, and each tool validates the live connection at call time. Two layers:

**1. MongoDB tools.** Native typed tools that expose MQL fields (`filter`, `projection`, `sort`, `pipeline`, `document`, `update`, …) as parameters and assemble the JSON-encoded request the driver already parses — the model never hand-authors MQL JSON:

| family | tools |
|--------|-------|
| SQL | `execute_query`, `execute_write`, `list_tables`, `describe_table` (unchanged) |
| MongoDB reads | `find_documents`, `aggregate_documents`, `count_documents`, `list_collections`, `describe_collection` |
| MongoDB writes (gated) | `insert_document`, `update_document`, `delete_document` |

`describe_collection` also surfaces indexes + the JSON Schema validator via the driver's existing `getCollectionInfo()`.

**2. The agent always knows SQL vs NoSQL, even across a mid-session connect/switch.** This is the key design point. An agent often connects its MCP client *before* a database is selected (typical for a troubleshooting session). The transport is stateless, so the server can't push `tools/list_changed` — a per-mode tool list would leave the agent holding a stale cached set and hitting cryptic "tool not found" after the DB is connected. Instead:

- **All tools are always advertised** (both families + a new `connection_info` tool), so `tools/list` never changes with DB type — there's no stale cache to invalidate.
- **Per-call mode guard**: a tool from the wrong family returns an *actionable* error naming the connected DB type and the tools that apply (e.g. *"This MCP connection is a MongoDB database… use find_documents/…; call connection_info / re-list / reconnect"*), so the agent self-corrects.
- **`connection_info`** (always available) reports the connected DB type, query language, write state, and recommended tools — the agent calls it to orient up front and to detect a switch.
- **Mode-aware instructions** name the currently-connected DB and point at `connection_info`.

This is client-agnostic (doesn't depend on a client honoring `tools/list_changed`) and self-correcting.

## Tests

New `server/src/mcp.test.ts` (9 tests, via the SDK in-memory transport):
- the toolset is invariant across SQL and Mongo connections
- `connection_info` reports DB type / query language / recommended tools
- a SQL tool on a Mongo connection (and vice-versa) returns an actionable mismatch error, **not** "tool not found"
- mode-aware instructions name the DB and point at `connection_info`
- `find_documents` assembles the correct `{collection, operation:'find', filter, limit}` JSON and passes schema through
- `insert_document` blocked when writes disabled, allowed when enabled
- `update_document` requires a `filter` or `id`

All 171 server tests pass; `mcp.ts` is type- and lint-clean. (3 pre-existing `tsc` errors in unrelated driver test files are untouched.)

## Known follow-ups (out of scope)
- `_id` ObjectId matching in `find`/`count` filters (string `_id` matched literally; relates to existing TODO #107). An `id` convenience param is exposed on update/delete.

### DIFF
--- a/server/src/mcp.test.ts
+++ b/server/src/mcp.test.ts
@@ -36,7 +36,13 @@ function mockMongoDriver(query = vi.fn()) {
   return driver;
 }
 
-describe('MCP server – driver-aware toolset', () => {
+function mockSqlDriver(query = vi.fn()) {
+  const driver = { queryMode: 'sql' as const, query };
+  vi.mocked(getDriver).mockReturnValue(driver as any);
+  return driver;
+}
+
+describe('MCP server – invariant toolset with per-call mode guards', () => {
   beforeEach(() => {
     vi.clearAllMocks();
     vi.mocked(isConnected).mockReturnValue(true);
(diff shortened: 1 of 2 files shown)


eugene-mobile:
@claude review this PR

eugene-mobile:
Thanks for the review — addressed in 73ab9a3.

**Bug — `aggregate_documents` unbounded fetch** ✅ Fixed. It now appends a `$limit` stage to the pipeline so MongoDB bounds the result set server-side instead of loading the whole aggregation into Node and slicing — same contract as `find_documents`. One tweak on the suggestion: I append `{$limit: cap + 1}` rather than `cap`, so the `truncated` flag is still accurate without a full scan. (Dropped `totalDocs` from the response, since reporting an exact total would require the unbounded scan we're avoiding.)

**Minor 1 — `describe_collection` optional chaining** ✅ Done: `await driver.getCollectionInfo?.(schema, collection) ?? null`.

**Minor 2 — test coverage gaps** ✅ Added MCP-layer tests for `aggregate_documents` (asserts the `$limit: cap+1` append + truncation), `count_documents`, `list_collections`, `describe_collection`, and `delete_document`. Suite is now 14 MCP tests / 176 server tests.

**Minor 3 — test cleanup** ✅ The invariant-toolset test now closes both client/server pairs.

**Minor 4 — empty `update: {}`** ⏭️ Intentionally skipped. MongoDB rejects an operator-less update at query time and that error already propagates cleanly through the tool's try/catch. A `.min(1)`/`refine` guard adds validator complexity for a case the DB already reports clearly, and risks false-rejecting edge cases — net negative.

`mcp.ts` is type- and lint-clean; all 176 server tests pass.

